PR #103, opened 2024-10-08
Add callback to "go to form" on double click event

jthompson-arcus:
Create double click event to "go to form" for overview table.


LDSamson:
Looks good. Two small comments: 
It would be nice to update the test description as well.

Since the feature is about browsing through the application, I think it is nice to preserve the old code (that you deleted) as a test for clicking the header widget.

I will push a change with these changes, let me know what you think.

tests/testthat/test-app_feature_01.R
```diff
@@ -54,6 +56,23 @@ describe(
...
+         output_names <- names(app$get_values(output = TRUE)$output)
+         app$expect_values(output = vector_select(output_names, exclude = "visit_figure"))
+         
+         ## Note (LSA): after clicking on 'graph' view again at this moment, an 
+         ## error will be shown instead of the expected figures. 
+         ## ('Error: incorrect height and width'). So far, this error could not 
+         ## be isolated and only occurred when running the script above 
+         ## non-interactively. Therefore, no further action is taken.
+         #app$view()
+         #app$set_inputs("sf_vital_signs-switch_view" = "graph")
```
LDSamson, line 75: @jthompson-arcus can you reproduce this? I cannot reproduce it interactively, only when testing with shinytest2, when placing a browser here and then run `app$set_inputs`. It's not related to this PR (I tested it and it already occurred in an older version as well)

jthompson-arcus: Yes, the error message is in the browser only (i.e. it doesn't show up in console). Seems to be connected to setting the width and height of the app driver somehow. Not really sure why it's happening or what to do to fix it and maintain current screen size for tests.

jthompson-arcus:
> Looks good. Two small comments: It would be nice to update the test description as well.
> 
> Since the feature is about browsing through the application, I think it is nice to preserve the old code (that you deleted) as a test for clicking the header widget.
> 
> I will push a change with these changes, let me know what you think.

Looks good to me.

LDSamson (approved):

Issue #113, opened 2024-10-16
Create Audit Trail in DB


jthompson-arcus:
Create a built in audit trail within the user database.

- [x] Review process
    - Add DB logging table
    - Add triggers to populate logging table
    - Modify processes to update instead of append
    - Process to update old DBs
    - Added tests
- [ ] Query process
    - Separate query table from comments
    - Add DB logging tables
    - Modify processes to update instead of append
    - Process to update old DBs
    - Added tests
- [ ] DB version
    - Function to help migrate between DB versions
    - Add checking in app for DB version
    - Add unit test to check function
- [ ] Simplify app process
    - Makes calls in the application more natural
    - Depend less on reactive objects, pull from DB

Issue #108, opened 2024-10-10
(Incorrect) error message shown when saving a review of a form


LDSamson:
To reproduce: 

Save a review of a form that previously was (partly) reviewed by someone else, or that was reviewed by the same user but has some new data points.

The following error will pop up:
‘An error occurred while trying to save the data. Try again or contact an administrator if the problem persists. Resynching in-memory data with database..’

However, the error is incorrect. The data is saved correctly. The problem here is that we check that the review is written in the database, but I incorrectly used `unique()` to select the unique row in the database. This returns multiple rows there are multiple reviewers of data in the same form (or if the reviewer is missing in some rows, but available in others):

https://github.com/openpharma/clinsight/blob/f32e09d92b0cc7893556ddedf875d4cd3fd3e249/R/mod_review_forms.R#L256-L273

Since the data is saved correctly, the issue does not have major impact, but the warning each time is of course undesirable.

PR #115, opened 2024-10-16
Add logging tables

jthompson-arcus:
Addresses #113 

To make this transition more manageable, I am going to break this up into multiple PRs. This PR will focus on the review table. I will also modify #113 to break the task up into smaller parts.

- [x] Add logging tables in DB creation.
- [x] Add trigger for updates on review table.
- [x] Modify process to update review table instead of append.
- [x] Create a new table to track DB version.
- [x] Create helper function to "upgrade" DBs
- [x] Add and modify tests to reflect new behaviors.


jthompson-arcus:
@LDSamson & @vanguyen317, code is still a bit rough since I'm just trying to show the implementation. Recent commits are to show how the database can be updated in the review process. The goal of this PR is to leave the current process intact and to only change the backend storage. Users shouldn't even notice a difference

jthompson-arcus:
@LDSamson, @aclark02-arcus, @dazheng-arcus,

Alright let me have it guys. I'm sure there will be changes requested. I think having lots of eyes and opinions on the PR would be best.

aclark02-arcus:
> DBI::dbGetQuery(con, "select * from all_review_data_ where id = 1")

Typo? It seems `all_review_data` is a table, but `all_review_data_` is not.

image

aclark02-arcus (commented):
Great work @jthompson-arcus! This is highly valuable work for the repo.

### Error messaging
I'm noticing a behavior that I can't quite explain. Wondering if you can help me. Below is a patient with only one record to review under Study Forms > Response:

image

After reviewing this record, notice that I got a message in the upper left hand corner. I then reviewed, un-reviewed, and reviewed the record several times. Just wanted to point it out and ask if it's a known issue? I'm actually less concerned about that message (though it should be resolved) but more concerned about what I found in the db (below).

#### Potential dupes

```r
DBI::dbGetQuery(con, "select * from all_review_data_log")
```
image

Notice that there are two `review_id`'s for (what I thought was) a single record. So, I decided to check out both `id`'s in `all_review_data`:

```r
DBI::dbGetQuery(con, "select * from all_review_data where id in (6221, 6222)")
```
image

The results show there may be duplicate entries for this record, with two different `id` fields. I didn't dig any deeper to see how that happened. Wondering if you can take a look?

### Timestamp confusion
One other comment - wondering if we could rename the timestamps in these tables to be more informative of what they represent - especially now that we have 3? I may botch this, but to my knowledge...
- `edit_date_time` is the time the data was last edited in the EDC?
- `timestamp` is the time the data was added to `all_review_data`?
- `dml_timestamp` is the time the data was logged?

Why is `timestamp` always equal to the previous record id's `dml_timestamp`?  If I need to perform some sort of audit on this data, I just want to know how all these data points are related / unrelated.

### Misc code questions
Also, I have a couple questions about the code below. None of them are problematic, but just thought I'd pose the question for my own understanding.

R/fct_SQLite.R
```diff
@@ -91,19 +91,70 @@ db_create <- function(
...
+                        "form_repeat", "item_name")
+  )
+  new_data <- list(
+    "db_synch_time"   = data.frame(synch_time = data_synch_time),
+    "db_version" = data.frame(version = "1.1")
   )
   con <- get_db_connection(db_path)
+  for(i in names(new_pk_data)){
+    cat("\nCreating new table: ", i,  "\n")
+    db_add_primary_key(con, i, new_pk_data[[i]], idx_pk_rows[[i]])
```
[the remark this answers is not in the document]
jthompson-arcus: I'm taking advantage of the fact that `idx_pk_rows[[i]]` will return `NULL` for the `query_data`. This PR isn't really handling the query data table, so I did not add an index constraint on it.

R/fct_SQLite.R
```diff
@@ -91,19 +91,70 @@ db_create <- function(
...
+        "review_id, edit_date_time, reviewed, comment, reviewer, timestamp, status, dml_type",
+      ")",
+      "VALUES(",
+        "NEW.id,",
+        "OLD.edit_date_time,",
+        "OLD.reviewed,",
+        "OLD.comment,",
+        "OLD.reviewer,",
+        "OLD.timestamp,",
+        "OLD.status,",
```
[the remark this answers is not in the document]
jthompson-arcus: They are keywords provided by SQL to differentiate the NEW and OLD records in the trigger.

R/mod_query_add.R
```diff
@@ -182,7 +182,7 @@ mod_query_add_server <- function(
       if(identical(new_query, query_in_db)){
         r$query_data <- dplyr::bind_rows(r$query_data, new_query)
       }
-      query_in_memory <- r$query_data[nrow(r$query_data), ]
+      query_in_memory <- r$query_data[nrow(r$query_data), -1]
```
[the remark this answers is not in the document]
jthompson-arcus: Yes, quick fix for the R CMD check. The tests for `all_review_data` was further modified to account for the new database table structure.

jthompson-arcus:
> #### Potential dupes
> ```r
> DBI::dbGetQuery(con, "select * from all_review_data_log")
> ```
> 
> image
> 
> Notice that there are two `review_id`'s for (what I thought was) a single record. So, I decided to check out both `id`'s in `all_review_data`:
> 
> ```r
> DBI::dbGetQuery(con, "select * from all_review_data where id in (6221, 6222)")
> ```
> 
> image
> The results show there may be duplicate entries for this record, with two different `id` fields. I didn't dig any deeper to see how that happened. Wondering if you can take a look?

@LDSamson Is this expected behavior in the table? I can see two different records when examining `clinsightfuldata`, but I would have expected a different `form_repeat` value for each record.

LDSamson:
> Great work @jthompson-arcus! This is highly valuable work for the repo.
> ### Error messaging
> 
> I'm noticing a behavior that I can't quite explain. Wondering if you can help me. Below is a patient with only one record to review under Study Forms > Response:
> 
> ![image]
> 
> After reviewing this record, notice that I got a message in the upper left hand corner. I then reviewed, un-reviewed, and reviewed the record several times. Just wanted to point it out and ask if it's a known issue? I stopped the app and relaunched using `load_and_run_app()` and the message seemed to go away. I'm actually less concerned about that message (though it should be resolved) but more concerned about what I found in the db after I ended my app session (keep reading below).
> #### Potential dupes
> 
> ```r
> DBI::dbGetQuery(con, "select * from all_review_data_log")
> ```
> 
> 
> Notice that there are two `review_id`'s for (what I thought was) a single record. So, I decided to check out both `id`'s in `all_review_data`:
> 
> ```r
> DBI::dbGetQuery(con, "select * from all_review_data where id in (6221, 6222)")
> ```
> 
> 
> 
> The results show there may be duplicate entries for this record, with two different `id` fields. I didn't dig any deeper to see how that happened. Wondering if you can take a look?
> ### Timestamp confusion
> 
> One other comment - wondering if we could rename the timestamps in these tables to be more informative of what they represent - especially now that we have 3? I may botch this, but to my knowledge...
> 
>     * `edit_date_time` is the time the data was last edited in the EDC?
> 
>     * `timestamp` is the time the data was added to `all_review_data`?
> 
>     * `dml_timestamp` is the time the data was logged?
> 
> 
> Why is `timestamp` always equal to the previous record id's `dml_timestamp`? If I need to perform some sort of audit on this data, I just want to know how all these data points are related / unrelated.
> ### Misc code questions
> 
> Also, I have a couple questions about the code below. None of them are problematic, but just thought I'd pose the question for my own understanding.

Hi both.
The combination of event_name, form_repeat, subject_id and item_group is a unique record which needs to be taken into account when creating the logging table.

I looked into it, and `Response` and `Responder` are two different items: Responder is dichotomous and Response is categorical, meaning that they can (and should) occur on the same page. See also item_value, which is different:
grafik

Regarding the error pop-up: it is a known bug, see the description here, and I am working on resolving it: 
https://github.com/openpharma/clinsight/issues/108#issue-2579108872

Furthermore: edit_date_time is indeed the time that an item was edited in the EDC. timestamp is when a user action was added to the database (for example a review or a query).

R/fct_SQLite.R
```diff
@@ -163,6 +214,27 @@ db_update <- function(
   cat("Finished updating review data\n")
 }
 
+db_upsert <- function(con, data, common_cols) {
```
LDSamson, line 217: Even for internal functions, it would be nice to have a roxygen description, which I think helps people who are less familiar with the code. What does it do, what are the inputs, the function arguments, and the expected output? I had to google what 'upsert' means (which probably shows he depth of my SQL knowledge ;) ).

R/mod_review_data_fct_helpers.R
```diff
@@ -88,29 +88,33 @@ update_review_data <- function(
...
+    dplyr::mutate(
+      timestamp = update_time,
+      reviewed = "No",
+      reviewer = NA_character_,
+      comment = "",
+      status = "updated"
+    )
+  
+  updated_data <- dplyr::bind_rows(new_records, updated_records)
+  updated_data$id <- NULL
```
LDSamson, line 117: Did you have a specific reason as to why this section needed rewriting? I will have to look what is going on here, with the triple anti_join, and come back to it.

R/mod_review_data_fct_helpers.R
```diff
@@ -88,29 +88,33 @@ update_review_data <- function(
...
-    dplyr::full_join(review_df, by = names(latest_review_data)) |> 
-    add_missing_columns(c("timestamp", "reviewed", "comment", "status"))
-  
-  df <- df |> 
-    tidyr::replace_na(list(timestamp = update_time, reviewed = "No", comment = "")) |> 
-    # now add status labels to the new rows:
+  # Grab the records not present in the review db table. Impute missing values.
+  new_records <- latest_review_data |> 
+    dplyr::anti_join(review_df, by = common_vars) |> 
+    add_missing_columns(c("timestamp", "reviewed", "comment", "status")) |>
```
LDSamson, line 94: I don't think add_missing_columns is needed if we are writing it this way, since the `mutate` call will add these columns anyways.

R/mod_query_add.R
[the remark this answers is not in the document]
jthompson-arcus: It will be changed in the future when/if the query table gets redesigned. I thought adding the primary key for now would be helpful, but that's all that has changed.

jthompson-arcus:
> Hi both. The combination of subject_id, event_name, form_repeat, and item_group should be a unique record.
> 
> I looked into it, and `Response` and `Responder` are two different items: Responder is dichotomous and Response is categorical, meaning that they can (and should) occur on the same page. See also item_value, which is different: grafik In other words: it is expected behavior that we get two different row ids for these events.

@LDSamson I think the problem here is that you say you expect two different row ids, but these records have the same subject_id, event_name, form_repeat, and item_group. They differ on other fields like item_name. So according to your first statement about unique records, we would only expect one. Unless of course we add item_name to the list of fields captured by all_review_records.

R/mod_review_data_fct_helpers.R
jthompson-arcus: This is a very good question. Initially I was needing to break the function up so I could perform both an INSERT and an UPDATE, but since I've moved towards using an UPSERT, I don't know if the changes are still needed. I will investigate.

LDSamson:
> but these records have the same subject_id, event_name, form_repeat, and item_group. They differ on other fields like item_name. S

My bad, item_name should be added here of course. Not sure if it needs to recorded if you have a unique identifier/key already.

jthompson-arcus:
> > but these records have the same subject_id, event_name, form_repeat, and item_group. They differ on other fields like item_name. S
> 
> My bad, item_name belongs to the variables identifying a unique row course. Not sure if it needs to recorded if you have a unique identifier/key already.

It will need to be recorded in the table. This is a pretty big change. Even if we didn't add the logging tables, we would need to fix this since uniqueness is being violated in the review table. This also means that migrating older tables is a lot more difficult and may not be possible.

LDSamson:
> > > but these records have the same subject_id, event_name, form_repeat, and item_group. They differ on other fields like item_name. S
> > 
> > 
> > My bad, item_name belongs to the variables identifying a unique row course. Not sure if it needs to recorded if you have a unique identifier/key already.
> 
> It will need to be recorded in the table. This is a pretty big change. Even if we didn't add the logging tables, we would need to fix this since uniqueness is being violated in the review table. This also means that migrating older tables is a lot more difficult and may not be possible.

So, the unique identifier/key that you created does not solve the uniqueness? The variables that I mentioned are called the `common_vars`, which are used several times, also in the functions that you created:   `common_vars = c("subject_id", "event_name", "item_group", "form_repeat", "item_name")`

R/fct_SQLite.R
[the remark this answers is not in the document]
aclark02-arcus: Learn something new everyday.

jthompson-arcus:
@aclark02-arcus I talked with @LDSamson and the potential dupes are indeed different records and is being taken into account in the all review data table. The `item_name` is getting pivoted wider into columns but the underlying data sees this as two records which individually could updated. For example, in this case the `Responder` field could get updated itself and have a different edit_date_time than `Response`.

R/mod_review_data_fct_helpers.R
jthompson-arcus: Have reverted back to `dev`.

R/fct_SQLite.R
```diff
@@ -91,17 +91,99 @@ db_create <- function(
...
+                     c(id = "INTEGER PRIMARY KEY AUTOINCREMENT", review_id = "INTEGER NOT NULL",
+                       edit_date_time = "CHAR", reviewed = "CHAR", comment = "CHAR", 
+                       reviewer = "CHAR", timestamp = "CHAR", status = "CHAR",
+                       dml_type = "CHAR NOT NULL", dml_timestamp = "DATETIME DEFAULT CURRENT_TIMESTAMP"))
+  rs <- DBI::dbSendStatement(con, paste(
+    "CREATE TRIGGER all_review_data_update_log_trigger",
+    "AFTER UPDATE ON all_review_data FOR EACH ROW",
+    "BEGIN",
+      "INSERT INTO all_review_data_log (",
+        "review_id, edit_date_time, reviewed, comment, reviewer, timestamp, status, dml_type",
```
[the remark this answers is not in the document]
LDSamson: Wait I think it is that the review_id identifies one 'review' that can contain multiple unique rows, and therefore the numbers are different, right?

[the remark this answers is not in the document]
jthompson-arcus: The `review_id` matches up with `all_review_data$id` (if I may use this notation). The `id` is the primary key for the logging table so that records can more easily be queried from that table. The logging table does not contain the index/unique columns. If you wanted to look at the log, you would need to join to `all_review_data` to get the index/unique columns. Storing those columns in both spaces wouldn't be very efficient.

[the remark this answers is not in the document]
jthompson-arcus: > Wait I think it is that the review_id identifies one 'review' that can contain multiple unique rows, and therefore the numbers are different, right?

Yes, review_id identifies which review the logging record is referencing.

R/fct_SQLite.R
```diff
@@ -369,3 +496,72 @@ db_get_review <- function(
       dplyr::as_tibble()
   })
 }
+
+update_db_version <- function(db_path, version = "1.1") {
```
LDSamson, line 500: Is it possible to create a test for this to verify if the result is as expected? It is a bit difficult for me to see if conversion is exactly as intended.

data-raw/internal_data.R
```diff
@@ -58,6 +58,16 @@ required_meta_cols <- c(
   "item_type"
   )
 
+idx_cols <- c(
```
LDSamson, line 61: A small description why it is needed and where it is used would be helpful here. What is 'idx'?

R/fct_SQLite.R
```diff
@@ -369,3 +496,72 @@ db_get_review <- function(
...
+  stopifnot(file.exists(db_path))
+  version <- match.arg(version)
+  temp_path <- withr::local_tempfile(fileext = ".sqlite")
+  file.copy(db_path, temp_path)
+  con <- get_db_connection(temp_path)
+  
+  current_version <- tryCatch({
+    DBI::dbGetQuery(con, "SELECT version FROM db_version") |> 
+      unlist(use.names = FALSE)}, error = \(e){""})
```
LDSamson, line 509: Maybe it would be helpful to extract this code as a small helper function db_get_version(), so that it is easy to check the db version.

R/fct_SQLite.R
```diff
@@ -369,3 +496,72 @@ db_get_review <- function(
...
+  if(identical(current_version, db_version)) return("Database up to date. No update needed")
```
LDSamson, line 510: It probably should only update if the version is currently missing, or if the version is at least smaller than 1.1, right? Just to protect us in the future

jthompson-arcus:
@LDSamson seems like you comments are mostly related to commenting and testing. Can those be incorporated into #135 instead of here?

R/fct_SQLite.R
```diff
@@ -91,17 +91,99 @@ db_create <- function(
...
+#'   table. Names will correspond to the table to apply the index constraint.
+#' @param other_data A named list of other data frames to add to the DB. Names
+#'   will correspond to the DB table names.
+#'
+#' @keywords internal
+db_add_tables <- function(con, pk_data, idx_cols, other_data) {
+  for(i in names(pk_data)){
     cat("\nCreating new table: ", i,  "\n")
-    DBI::dbWriteTable(con, i, new_data[[i]])
+    db_add_primary_key(con, i, pk_data[[i]], idx_cols[[i]])
```
LDSamson, line 124: It is confusing to have a internal data object named `idx_cols`, which is a character vector, and also this object, with the same name, which is supposed to be a list. Is this actually doing the correct thing here? 

I would highly suggest to use different names for these to avoid confusion.

R/fct_SQLite.R
```diff
@@ -91,17 +91,99 @@ db_create <- function(
...
+#' @param con A DBI Connection to the SQLite DB
+#' @param pk_data A named list of data frames to add a primary key field to DB
+#'   table. Names will correspond to the DB table names.
+#' @param idx_cols A named list of the fields defining unique records for a
+#'   table. Names will correspond to the table to apply the index constraint.
+#' @param other_data A named list of other data frames to add to the DB. Names
+#'   will correspond to the DB table names.
+#'
+#' @keywords internal
+db_add_tables <- function(con, pk_data, idx_cols, other_data) {
```
LDSamson, line 121: The argument `idx_cols` is different here from the internal object named `idx_cols` (=character vector). Is the correct one used in the function?

I would suggest to use different names, this is confusing for me at least.

R/fct_SQLite.R
jthompson-arcus: I don't really think adding a function would simplify anything. I just copied how you handled this in `db_update()`.

R/fct_SQLite.R
jthompson-arcus: If we need this type of check in the future, we can add it then.

data-raw/internal_data.R
jthompson-arcus: Done.

R/fct_SQLite.R
jthompson-arcus: Done.

R/fct_SQLite.R
LDSamson: @jthompson-arcus as discussed on Teams today: we will look into how to test a database update with the function `update_db_version` in a future PR

LDSamson (approved):
Looks good. I tested it with a database of ours that is used in production, and I could use the updated database locally with this version of clinsight without any problems.

For future improvements, as discussed, it would be great if we can add tests for a database update. Furthermore, the id row should ideally never change because it connects the review_data table with the logging table; @jthompson-arcus agreed to check if this is covered in the unit tests and if not, add such a test in a follow-up PR.

Issue #107, opened 2024-10-10
create_table.default requires `form_repeat` in the `keep_vars` argument


LDSamson:
Currently it is impossible to create a new `common_forms` table without creating a new S3 method for it, since create.table_default will remove the `form_repeat` variable, which is a mandatory variable for `common_forms` since the changes in PR #92 were implemented. 

https://github.com/openpharma/clinsight/blob/92d4659282109eafe19d109092d0691a9b2568c1/R/fct_appdata_summary_tables.R#L156

and: 

https://github.com/openpharma/clinsight/blob/92d4659282109eafe19d109092d0691a9b2568c1/R/fct_appdata_summary_tables.R#L177

PR #118, opened 2024-10-23
Extend `create_table()`

jthompson-arcus:
Addresses #107


R/fct_appdata.R
```diff
@@ -233,7 +233,17 @@ get_appdata <-  function(
...
+  form_types <- lapply(
+    c("common_forms", "study_forms", "general"),
+    \(x) {
+      lapply(metadata[x], \(y) {
+        cbind(form_type = x, item_group = unique(y$item_group))
+      }) |> 
+        do.call(what = rbind)
+    }) |> 
+    do.call(what = rbind) |> 
+    as.data.frame()
```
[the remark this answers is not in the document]
jthompson-arcus: I'm not opposed. I think if we come up with more utility for the field that could make sense. I simplified the creation of the object a little bit by using `Map()` instead of nested `lapply()`s. I think it looks a bit less awkward now.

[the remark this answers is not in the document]
LDSamson: If you are not opposed to it, maybe it is better to add it to the metadata$items_expanded table. 

By doing so we don't even need to use `lapply` or `Map` in `get_appdata`.

In addition, I plan to remove the tabs `common_forms`, `general`, and `study_forms` from the metadata object in the future since it is redundant information (all this information is also in `items_expanded`)

R/fct_tables.R
```diff
@@ -176,6 +176,23 @@ create_table.general <- function(
 }
 
 
+#' Create Default 'Common Events' Table
+#'
+#' @export
+#' @inherit create_table.default 
+create_table.common_events <- function(
```
LDSamson, line 183: I think the function should be named `create_table.common_forms`, right?

R/fct_appdata.R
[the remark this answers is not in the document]
jthompson-arcus: `form_type` column has been added to `items_expanded`

R/fct_tables.R
jthompson-arcus: Oopsies

jthompson-arcus:
@LDSamson we good here?

R/fct_appdata.R
```diff
@@ -267,9 +267,15 @@ get_appdata <-  function(
...
+    form_type_x <- unique(with(var_levels, form_type[item_group == item_group_x]))
+    if(length(form_type_x) != 1) stop(
+      "form_type consists of multipe elements which is not allowed: ", 
+      form_type_x
+    )
+    class(x) <- c(form_type_x, class(x))
     tableclass <- simplify_string(item_group_x)
     if(tableclass %in% tableclasses){
-      class(x) <- c(tableclass, class(x)) 
+      class(x) <- unique(c(tableclass, class(x)))
```
LDSamson, line 278: I'm not against this, but I was just wondering if there was a specific reason to wrap this in `unique`? Did you encounter a problem without it?

R/fct_appdata.R
```diff
@@ -312,7 +318,7 @@ get_appdata <-  function(
         )
       ) |> 
       dplyr::ungroup() 
-    class(df) <- c("continuous", class(df))
+    class(df) <- unique(c("continuous", class(x)))
```
LDSamson, line 321: See previous comment

R/fct_appdata.R
```diff
@@ -267,9 +267,15 @@ get_appdata <-  function(
       "item_group consists of multipe elements which is not allowed: ", 
       item_group_x
     )
+    form_type_x <- unique(with(var_levels, form_type[item_group == item_group_x]))
```
LDSamson, line 270: Maybe it is good to wrap form_type_x also in `simplify_string`. That way the user is protected against some hard to detect changes in the names (for example, accidentally adding a leading or trailing white space in the table's name in the Excel file)

LDSamson (commented):
I have one small suggestion, and a question out of curiosity. Otherwise, it looks all good to me.

R/fct_appdata.R
jthompson-arcus: Just to be tidy. Otherwise there are/can be duplicates in the class of an object.

LDSamson:
Thanks! looks good

PR #123, opened 2024-11-01
Use base64 encoding for usernames in http headers (shinyproxy specific)

### LINKED ISSUE #125: Using non-ascii signs in user names will cause a crash in `shinyproxy` and thus in `clinsight`
See issue reported [here](https://github.com/openanalytics/shinyproxy/issues/533).

Root cause is that apparently only ascii signs can be transferred in HTTP headers, and a workaround is to always use base64 encoding to transfer user names.

LDSamson:
Closes #125. HTTP headers can only contain ascii signs, which causes issues when using names with less common names. Therefore, a solution is that user names are base64 encoded by the application serving the app (usually `shinyproxy`), and then decoded by clinsight once done.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -46,6 +46,7 @@ Imports:
     tools,
     withr
 Suggests: 
+    base64enc,
     kableExtra,
     knitr,
     pkgload,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: clinsight
 Title: ClinSight
-Version: 0.1.0.9004
+Version: 0.1.0.9006
 Authors@R: c(
     person("Leonard Daniël", "Samson", , "lsamson@gcp-service.com", role = c("cre", "aut"),
            comment = c(ORCID = "0000-0002-6252-7639")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,9 +1,15 @@
 # clinsight (development version)
 
+## Changed 
+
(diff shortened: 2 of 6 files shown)


jthompson-arcus:
Not really familiar with what's going on here. Are you encoding the values into base 64 in the deployment and this PR is just decoding/checking those values?

jthompson-arcus:
Is the first test an example of a user name that would have caused issues previously?

LDSamson:
Yes exactly, to both of your questions. See the news description. I now ensure that all user names are base 64 encoded in shinyproxy, and then they need to be decoded in the app. This PR will do just that (decoding the user names). The provided user name in the test indeed previously caused issues.

DESCRIPTION
```diff
@@ -46,6 +46,7 @@ Imports:
     tools,
     withr
 Suggests: 
+    base64enc,
```
jthompson-arcus, line 49: I have mixed feelings about this but since we are including `base64enc` in the suggests, I think `decode_base64()` should at the very list check if the packaged has been installed. (I have mixed feelings because it is a dependency for other packages, so I can't really think of a scenario where it wouldn't be available.)

LDSamson: It is a function that will only run in production. It should be installed already since the package `bslib` depends on it. What other reservations do you have for adding it in suggests, other than checking in `decode_base64` a check if the package is installed?

I will add the check mentioned by you in the function.

jthompson-arcus (approved):
LGTM

Issue #96, opened 2024-10-01
Allow for excel / csv export of listings


aclark02-arcus:
A lot of users want to download the data shown in the app. Wondering if there are any objections to adding a csv or excel button to all the `DT` tables?

aclark02-arcus:
Oh whoops. Re-opening this one until PR merged on `dev`.

LDSamson:
@aclark02-arcus can you create a PR on dev for this when you have time? I think it's nice to include there as well

PR #124, opened 2024-11-04
Give users ability to re-order columns

aclark02-arcus:
I don't think I ever opened an issue for this one here, but this is probably the most requested feature to `clinsight` for us. Kind of related to #96 because it reduces reliance on exporting to excel, when you can control the variable order in the app. I was too lazy to record a `gif`, but give it a run and you can now drag and drop columns for any table.

The only known issue with this extension is listed here: https://github.com/rstudio/DT/issues/250

### DIFF
--- a/R/fct_data_helpers.R
+++ b/R/fct_data_helpers.R
@@ -542,10 +542,10 @@ datatable_custom <- function(
     rename_vars = NULL, 
     title = NULL, 
     selection = "single",
-    extensions = "Scroller",
+    extensions = c("Scroller", "ColReorder"),
     plugins = "scrollResize",
     dom = "fti",
-    options = list(),
+    options = list(colReorder = TRUE),
     ...
     ){
   stopifnot(is.data.frame(data))
--- a/R/fct_data_helpers.R
+++ b/R/fct_data_helpers.R
@@ -528,6 +528,7 @@ add_missing_columns <- function(
 #'     * `deferREnder = TRUE`
 #'     * `scrollResize = TRUE`
 #'     * `scrollCollapse = TRUE`
+#'     * `colReorder = TRUE`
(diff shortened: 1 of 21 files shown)


LDSamson:
Looks okay to me. Probably nice to give users some flexibility here. 

The preference will of course not be saved. Does everyone in your team have their own column preferences? If so: the column order is usually set based on the order of the variables in the `metadata.xlsx` file.

R/fct_data_helpers.R
```diff
@@ -542,10 +542,10 @@ datatable_custom <- function(
     rename_vars = NULL, 
     title = NULL, 
     selection = "single",
-    extensions = "Scroller",
+    extensions = c("Scroller", "ColReorder"),
     plugins = "scrollResize",
     dom = "fti",
-    options = list(),
+    options = list(colReorder = TRUE),
     ...
```
LDSamson, line 549: @aclark02-arcus I think it is probably better to add this to the list of modifyable defaults and add it to the function's description. @jthompson-arcus what do you think? In addition, can you also update the package documentation and include it in the PR?

LDSamson (changes requested):

R/fct_data_helpers.R
jthompson-arcus: Yes, with how this function is currently setup, default options should be included in `default_opts`, not the options argument.

LDSamson (changes requested):
Can you review and fix the test errors? otherwise I think it looks okay.

The tests capture the datatable output in their snapshot, I think that is why it is failing.

aclark02-arcus:
Hi @LDSamson, got the coveted ✅

LDSamson (approved):
Looks good to me

PR #134, opened 2024-11-07
Fix database inconsistency warnings when saving a review of a form

### LINKED ISSUE #108: (Incorrect) error message shown when saving a review of a form

LDSamson:
Fixes #108.

Verifications are now improved when saving a form review. Now the verifications and the update of the in-memory review data frame take into account that the review state might differ per item within a form, and thus that not all items need an update of the review state when saving a review.


jthompson-arcus (commented):
LGTM! No notes.

@LDSamson did you want to also open a version of this PR to `main` since you are seeing this bug in production?

jthompson-arcus:
Actually it might be better to redo the PR as a direct merge to `main`, then we can merge `main` back into `dev`.

LDSamson:
As discussed, I created a PR for the main branch that will fix this issue (#138 ). I will close this PR for now.  We can probably use [this commit](https://github.com/openpharma/clinsight/pull/134/commits/5c32c54ec01fd1c58484133a07add8d7403ca9e3) to fix any merge issues with dev later on.

PR #135, opened 2024-11-08
Simplify review process

jthompson-arcus:
Addresses part of #113. Follows PR #115.


jthompson-arcus:
This PR has two goals:

1. Convince ourselves that `db_slice_rows()` is not needed to review forms any more.
2. Remove `db_slice_rows()` from the review process.

jthompson-arcus:
@LDSamson WRT our discussion about whether the ID field can be updated, It was indirectly captured via a snapshot that the ID did not get updated [here](https://github.com/openpharma/clinsight/blob/f271502c50774606afe0d849784d2e8221983d79/tests/testthat/test-fct_SQLite.R#L181-L197). I have also added a trigger to this PR that will literally cause a failure if that field is attempted to be updated.

R/fct_SQLite.R
```diff
@@ -164,6 +164,16 @@ db_add_log <- function(con) {
...
+  # This will trigger before any UPDATEs happen on all_review_data. Instead of
+  # allowing 'id' to be updated, it will throw an error.
+  rs <- DBI::dbSendStatement(con, paste(
+    "CREATE TRIGGER all_review_data_id_update_trigger",
+    "BEFORE UPDATE OF id ON all_review_data",
+    "BEGIN",
+    "SELECT RAISE(FAIL, 'all_review_data.id is read only');",
+    "END"
+  ))
+  DBI::dbClearResult(rs)
```
LDSamson, line 176: I tried to test this behavior with a test database. Could it be that these are not updated yet? Also, is it possible to capture this in a test? It is quite important that this does not go wrong.

R/fct_db_collect_data.R
```diff
@@ -82,6 +82,15 @@ db_slice_rows <- function(
   dplyr::slice_tail(df, n = 1, by = dplyr::all_of(c(group_vars, slice_vars)))
 }
 
+db_get_table <- function(db_path, db_table = "all_review_data") {
+  stopifnot(is.character(db_path))
+  stopifnot(is.character(db_table))
+  
+  con <- get_db_connection(db_path)
+  dplyr::tbl(con, db_table) |> 
+    dplyr::collect()
```
LDSamson, line 91: Following up on our discussion last week: I did some quick testing and using a SQL query directly is 2x (and up to 10x) faster. It's still only about milliseconds but maybe it is better to switch to use something like below for functions that we are calling within the application non-interactively.

```
con <- get_db_connection(db_path)
sql <- "SELECT * FROM ?db_table;"
query <- DBI::sqlInterpolate(con, sql, db_table = db_table[1])
DBI::dbGetQuery(con, query)
```

R/run_app.R
```diff
@@ -67,6 +67,8 @@ run_app <- function(
     warning("No user database found. New database will be created")
     db_create(get_review_data(data), db_path = user_db)
   } else{
+    stopifnot("user_db version is not update to date" =
```
LDSamson, line 70: Typo in the stop message ;)

tests/testthat/test-mod_review_forms.R
```diff
@@ -129,8 +129,8 @@ describe(
             r_id <- with(db_reviewdata, id[subject_id == "885"])
             expect_equal(with(db_reviewlogdata, comment[review_id == r_id]), c("", ""))
             expect_equal(with(db_reviewlogdata, reviewed[review_id == r_id]), c("No", "Yes"))
-            expect_equal(r$review_data, db_slice_rows(db_path))
-            expect_equal(r$review_data, dplyr::arrange(db_reviewdata, edit_date_time))
```
LDSamson, line 133: Nice that `dplyr::arrange` is not needed anymore

R/fct_SQLite.R
```diff
@@ -325,13 +335,11 @@ db_save_review <- function(
     dplyr::inner_join(dplyr::tbl(db_con, "row_ids"), by = review_by) |> 
     # Filter below prevents unnecessarily overwriting the review status in forms   
     # with mixed reviewed status (due to an edit by the investigators). 
-    dplyr::filter(reviewed != new_review_state) |>
```
LDSamson, line 328: This can be removed? Thus, identification of which rows within a form needs an update is completely done in the SQL statement in this function now? I will double check later.

LDSamson (commented):
@jthompson-arcus I have a few comments, see below.

R/fct_SQLite.R
jthompson-arcus: I did not realize this was part of the review process until reviewing your latest PR fixing the errant error message. I actually fixed it the follow-up PR. I will add this line back in here.

R/fct_SQLite.R
jthompson-arcus: Should this behavior be extended to all `idx_cols`?

LDSamson: Good question. I feel like the id column is the most important since it links the review data to the audit/log table. 

I am trying to thing if there would be a moment that one of the idx_cols/common_vars gets updated. 
What would happen if this occurs for the event_name (e.g. from `visit 1` to `visit 2`, although I think that would be rare)? Would that count as a completely new row with a new id?

jthompson-arcus: Well with the construction as it stands now (and how it functioned previously I believe) this would get input to the review data base as a new record and a warning would be triggered indicating "items were not found in the updated dataset". See `update_review_data()`.

R/fct_db_collect_data.R
jthompson-arcus: Updated

R/run_app.R
jthompson-arcus: Fixed

R/fct_SQLite.R
jthompson-arcus: Added back in

LDSamson:
@jthompson-arcus I think it looks okay as far as I can see but it is a bit difficult to see the differences between the changes I made (in the currently open PR on main) and this one, for the tests. I will have a closer look tomorrow.

LDSamson (approved):
> @LDSamson let me know what you think of these changes. I think adding a test could be good but will need to rework `db_add_primary_key()` and `db_add_log()` to not be so `all_review_data` focused.

It would be great to add tests indeed since this is a vital part of functioning of the application. But it looks good and the tests can be added in a follow-up PR, it all looks like it is working as intended now.

Issue #99, opened 2024-10-01
Allow review of singular records / rows


aclark02-arcus:
The way our current software works is each record in the listing has to be reviewed, instead of the entire form / domain at once. Pilot users cited instances where they'll spend hours reviewing a single AE, so reviewing the entire AE form seems a bit overwhelming to them, assuming there are several AEs to review.

LDSamson:
I think this can be closed now that row review is implemented in the `dev` branch

PR #137, opened 2024-11-13
Review by records instead of `subject_id` & `form_id` "rv_row"

jthompson-arcus:
Still lead up to addressing #99. Review process needs to depend on records being reviewed, not the `review_row` it currently does.


R/fct_SQLite.R
```diff
@@ -372,7 +356,9 @@ db_save_review <- function(
     "FROM",
     "row_updates",
     "WHERE",
-    sprintf("%s.id = row_updates.id", tables)
+    sprintf("%s.id = row_updates.id", tables),
+    "AND",
+    sprintf("%s.reviewed <> row_updates.reviewed", tables)
```
jthompson-arcus, line 361: Maintains current structure where updates only occur when reviewed status is being changed/updated.

R/fct_SQLite.R
```diff
@@ -498,28 +483,26 @@ db_get_query <- function(
...
 #' 
 db_get_review <- function(
     db_path, 
-    subject = review_row$subject_id, 
-    form = review_row$item_group,
-    db_table = "all_review_data",
-    slice_vars = c("timestamp", "edit_date_time"),
-    group_vars = c("subject_id", "event_name", "item_group",
-                   "form_repeat", "item_name")
+    ids = review_row$id,
```
jthompson-arcus, line 491: Substantial change that's open for discussion. Makes it harder to see which records are being grabbed in the tests.

jthompson-arcus: NM. I generalized the function.

R/fct_SQLite.R
```diff
@@ -319,51 +319,35 @@ db_upsert <- function(con, data, idx_cols) {
...
+    tables = c("all_review_data")
 ){
-  stopifnot(is.data.frame(rv_row))
-  if(nrow(rv_row) != 1){
-    warning("multiple rows detected to save in database. Only the first row will be selected.")
-    rv_row <- rv_row[1, ]
+  stopifnot(is.data.frame(rv_records))
+  if (any(duplicated(rv_records[["id"]]))) {
+    warning("duplicate records detected to save in database. Only the first will be selected.")
+    rv_records[!duplicated(rv_records[["id"]]),]
```
LDSamson, line 340: I think you should save `rv_records` again (`rv_records <- ...`) in order to make this filter last

R/mod_review_forms.R
```diff
@@ -136,8 +132,8 @@ mod_review_forms_server <- function(
         # it will give a warning. This would be rare since it would mean a datapoint with the same edit date-time was reviewed but another one was not. 
         # probably better to use defensive coding here to ensure the app does not crash in that case. However we need to define which review status we need to select
         # in this case get the reviewed = "No"
-        review_status  <- unique(review_data_active()$reviewed)
-        review_comment <- unique(review_data_active()$comment)
+        review_status <- with(review_data_active(), reviewed[edit_date_time == max(as.POSIXct(edit_date_time))]) |> unique()
+        review_comment <- with(review_data_active(), comment[edit_date_time == max(as.POSIXct(edit_date_time))]) |> unique()
```
LDSamson, line 136: If we are going to review by item, then items can have a different review status with the same edit_date_time and this selection can contain multiple `review_status` and review comments

R/mod_review_forms.R
```diff
@@ -231,8 +227,8 @@ mod_review_forms_server <- function(
       review_save_error(FALSE)
       golem::cat_dev("Save review status reviewed:", input$form_reviewed, "\n")
       
-      review_row <- review_data_active() |> 
-        dplyr::distinct(subject_id, item_group) |> 
+      review_records <- review_data_active() |> 
+        dplyr::distinct(id) |>
```
LDSamson, line 231: Just wondering why to use `distinct()`. I thought `id` can never be duplicated with how it is created?

R/mod_review_forms.R
```diff
@@ -242,45 +238,40 @@ mod_review_forms_server <- function(
         ) 
       
       golem::cat_dev("review row to add:\n")
```
LDSamson, line 240: cans be changed to `review records to add`

R/mod_review_forms.R
```diff
@@ -110,14 +110,10 @@ mod_review_forms_server <- function(
...
-      df <- r$review_data |>
+      r$review_data |>
         dplyr::filter(subject_id == r$subject_id, 
                       item_group == active_form()) |> 
-        dplyr::distinct(subject_id, item_group, edit_date_time, reviewed, comment, status)
-      #!! below selects the latest edit_date_time; usually only one row will remain by then since there are no items displayed here.
-      if(nrow(df)== 0) return(df)
-      df |> 
-        dplyr::filter(edit_date_time == max(as.POSIXct(edit_date_time)))
+        dplyr::select(id, dplyr::all_of(idx_cols), edit_date_time, reviewed, comment, status)
```
LDSamson, line 116: Is this select statement still needed? Or can we just leave it out?

R/fct_SQLite.R
```diff
@@ -488,38 +474,38 @@ db_get_query <- function(
...
+    fields <- ...names()
+    sql <- paste("SELECT * FROM ?db_table WHERE", paste0(fields, " = $", fields, collapse = " AND "))
+    query <- DBI::sqlInterpolate(con, sql, db_table = db_table[1])
+    rs <- DBI::dbSendQuery(con, query)
+    DBI::dbBind(rs, params = data.frame(...))
+    df <- 
+      DBI::dbFetch(rs) |> 
       dplyr::as_tibble()
+    DBI::dbClearResult(rs)
+    df
```
LDSamson, line 508: Clever construction. I did not know the `...names()` notation before. However, I am wondering if that is not too flexible in our use case. What if we only select based on 'id'? It would make this function easier to check and to maintain. 

Furthermore, currently the function requires at least one non-defined column in `...` to run (otherwise you'll get an error: `Error: near "=": syntax error`). It would be nice to throw an informative error before attempting this, and to describe that at least one argument is needed.

R/fct_SQLite.R
LDSamson: Note that the `tables` argument could contain multiple names to update in the database, but not anymore since your last changes. Maybe change it to 'table' and let it now error when multiple tables are entered (and update the description).

tests/testthat/test-mod_review_forms.R
```diff
@@ -230,7 +230,7 @@ describe(
         expect_true(app$get_js("document.getElementById('test-review_comment').disabled;"))
         
         # review status and reviewer is saved as expected
-        saved_review_row <- db_get_review(temp_path, subject = "885", form = "Adverse events")
+        saved_review_row <- db_get_review(temp_path, subject_id = "885", item_group = "Adverse events")
```
LDSamson, line 233: Was the reason of making `db_get_review` general (and not only using `id`) so that this test can stay largely the same?

tests/testthat/test-mod_review_forms.R
```diff
@@ -279,16 +279,12 @@ describe(
...
-              ) |>
-              dplyr::select(subject_id, item_group, edit_date_time, reviewed, comment, status)
+            review_data_active(),
+            dplyr::filter(r$review_data, subject_id == "885", item_group == "Adverse events") |> 
+              dplyr::select(id, dplyr::all_of(idx_cols), edit_date_time, reviewed, comment, status)
           )
-          expect_equal(review_data_active()$item_group, "Adverse events")
-          expect_equal(nrow(review_data_active()), 1)
+          expect_equal(review_data_active()$item_group, c("Adverse events", "Adverse events"))
+          expect_equal(nrow(review_data_active()), 2)
```
LDSamson, line 287: The scenario description should be updated as well since we expect two rows now

R/fct_SQLite.R
jthompson-arcus: So I did think about checking that `...` corresponded to the column names for the table in the DB, but I don't really want to run that check in the application itself. I suppose I could add an argument `check_names = TRUE` and we could set this as false in the app.

R/mod_review_forms.R
jthompson-arcus: Yes, but that's not in this PR. Still maintaining current functionality for the time being.

R/fct_SQLite.R
jthompson-arcus: Oops

R/fct_SQLite.R
jthompson-arcus: For the time being I will relax the restrictions where if `...` is left empty the function will return the whole table.

LDSamson: With my second comment, I meant adding something like a simple `stopifnot()` with an informative message, testing if `...names()` is not `NULL`. The SQL statement will already error if any name is incorrect, and that gives a fairly clear message I think (after a quick check).

R/mod_review_forms.R
jthompson-arcus: I think this just got left in when I was updating code and didn't get around to removing it.

R/mod_review_forms.R
jthompson-arcus: I will remove it.

R/fct_SQLite.R
```diff
@@ -488,38 +475,48 @@ db_get_query <- function(
...
+        warning("No arguments passed in `...`. Returning full data table.")
+      conditionals <- "true"
+    } else {
+      conditionals <- paste0(fields, " = $", fields, collapse = " AND ")
+    }
+    sql <- paste("SELECT * FROM ?db_table WHERE", conditionals)
+    query <- DBI::sqlInterpolate(con, sql, db_table = db_table[1])
+    rs <- DBI::dbSendQuery(con, query)
+    if (!is.null(fields))
+      DBI::dbBind(rs, params = data.frame(...))
```
LDSamson, line 514: For my understanding: why is it that parameters must have the same the same length and thus you need to use `data.frame` instead of list? Shouldn't it also work with different length objects?

If the parameters indeed cannot be of different length, isn't it better to let the function error before sending a SQL query in such a case? Otherwise the results might not be cleared properly

R/fct_SQLite.R
```diff
@@ -488,38 +475,48 @@ db_get_query <- function(
...
-    DBI::dbGetQuery(con, query) |> 
-      db_slice_rows(slice_vars = slice_vars, group_vars = group_vars) |> 
+    fields <- ...names()
+    if (is.null(fields)) {
+      if (...length() > 0)
+        warning("Unnamed arguments passed in `...`. Returning full data table.")
+      else
+        warning("No arguments passed in `...`. Returning full data table.")
+      conditionals <- "true"
+    } else {
```
LDSamson, line 507: Can we capture this in a test? Should be quite straightforward I think!

R/mod_review_forms.R
```diff
@@ -231,8 +226,7 @@ mod_review_forms_server <- function(
       review_save_error(FALSE)
       golem::cat_dev("Save review status reviewed:", input$form_reviewed, "\n")
       
-      review_row <- review_data_active() |> 
-        dplyr::distinct(subject_id, item_group) |> 
+      review_records <- review_data_active() |>
```
LDSamson, line 229: We could still only select the id column so that review_records only contains the minimum necessary information for the update. Something like below

```
review_records <- review_data_active()["id"] |>
    ...
```

R/mod_review_forms.R
```diff
@@ -241,46 +235,39 @@ mod_review_forms_server <- function(
...
+          dplyr::rows_update(review_records, by = "id")
       }
       
-      updated_items_memory <- sort(with(r$review_data, item_name[
-        reviewer == review_row$reviewer[1] & timestamp == review_row$timestamp[1]
-      ]))
-      updated_items_db <- sort(updated_rows_db$item_name)
+      updated_items_memory <- review_records |> 
+        dplyr::left_join(r$review_data, by = "id", suffix = c("", ".y")) |> 
+        dplyr::select(dplyr::all_of(names(review_records))) |>
```
LDSamson, line 265: I am a bit confused with this. 
We want to check here that `r$review_data` contains the correct updated rows after the `rows_update` step above.

Why not just filter `r$review_data` on all the ids in `review_records`? Would that be a problem? 
```
r$review_data[r$review_data$id %in% review_records$id, ]
```

Furthermore, I think in the current setup, the review columns from r$review_data (status/comment/etc) are deleted and we are just checking the ones from `review_records` by using `suffix = c("", "y.")`. Thus, we are not checking what is actually stored in memory (in `r$review_data`).

Am I missing something here?

Also, minor: the name is now a bit confusing for me: before I was simply comparing a vector of items, hence the name `updated_items_memory`.

LDSamson (changes requested):
Some minor remarks. One change confuses me a bit though, see comments

R/mod_review_forms.R
jthompson-arcus: Yes, I seem to have missed the plot when updating this code.

LDSamson: I've added some small fixes (e.g. selected columns were not the same and therefore the comparison failed)

LDSamson:
@jthompson-arcus I wanted to make some minor changes and then merge this branch, but I realized that bug #108 was reintroduced in this PR when testing interactively, and our tests did not capture it properly. 
image

I updated the tests so that they would now fail as expected, and have a look at it at a later moment.

R/fct_SQLite.R
LDSamson: @jthompson-arcus I removed the export tag, and adjusted it so that it will now error before sending a SQL query if the `...` arguments cannot be coerced to a data frame. I guess that is the easiest option for now; we can improve it with you aforementioned suggestion once we want to export it again

tests/testthat/_snaps/mod_review_forms/test-mod_review_forms-003.json
```diff
@@ -17,6 +17,6 @@
     }
   },
   "export": {
-
+    "test-review_save_error": false
   }
```
LDSamson, line 21: This test is failing but I changed the snapshot manually to what it should show, for now

R/mod_review_forms.R
```diff
@@ -241,46 +235,40 @@ mod_review_forms_server <- function(
...
+        ) |> 
+        dplyr::select(dplyr::all_of(names(review_records)))
       
-      review_row_db <- unique(updated_rows_db[names(review_row)])
-      if(identical(review_row_db, review_row)){
+      review_records_db <- updated_rows_db |> 
+        # Within a form, only items with a changed review state are updated and 
+        # contain the new (current) time stamp. 
+        dplyr::filter(timestamp == review_records$timestamp[1])
+      if(isTRUE(all.equal(review_records_db, review_records, check.attributes = FALSE))){
```
LDSamson, line 257: The issue is that this statement now fails: the data frame `review_records` contains all possible item updates, but in the database only records with a changed review status will be updated. Thus, this statement does not always return `TRUE`

LDSamson:
@jthompson-arcus I think I fixed it. Let me know what you think of it

jthompson-arcus:
> @jthompson-arcus I think I fixed it. Let me know what you think of it

@LDSamson no real notes. I think it will be changing once again in #146 but I think this is good for now.

LDSamson (approved):

PR #139, opened 2024-11-19
Fix codecov uploads

LDSamson:
Fixes `codecov` uploads. Updates the workflow to the latest available one.


LDSamson:
@jthompson-arcus note that the old workflow might even work now, because James uploaded an organization-wide `CODECOV_TOKEN`. But this is according to the latest example workflow, probably more future proof, using https://github.com/r-lib/actions/blob/v2-branch/examples/test-coverage.yaml but with codecov-action V5: https://github.com/marketplace/actions/codecov#v5-release

jthompson-arcus (approved):
Love it!

PR #145, opened 2024-11-21
Use `renv` for GHA

jthompson-arcus:


LDSamson:
That interaction is a bit confusing for me as well. I think `setup-r-dependencies` does not use the `renv.lock` file to install exact versions. But I think it uses the standard local `renv` library storage, so it will install from there first if a package is available (at least that is my guess).

Better to use one I think indeed. I used `setup-r-dependencies ` in addition to `setup-renv` as a (inefficient) way to install additional dependencies (e.g. `covr`), but these are now included in the `full` renv lockfile profile so what you propose here should be better I think.

Also, it's probably better to merge GHA workflow changes to main as well. No need to update the version I think since the actual code itself is not changing. What do you think?

jthompson-arcus:
It is also not hard to just install a package if it's needed in the GHA. To me it makes more sense to be testing the package in the theoretical environment it is designed to be ran in. We can of course reconsider if we want to release on CRAN or we expect people to be downloading directly from GitHub as opposed to forking/cloning the project.

I'm not so concerned about merging it to `main` by itself. We can but I don't really see it as a value add. More important on the development side of things.

LDSamson (approved):

Issue #22, opened 2024-05-29
Allow for alternative review workflow - by form


aclark02-arcus:
For larger phase III studies, there can be thousands of patients, which makes it difficult  /impractical for medical monitors to browse patients-by-patient through the data. A seemingly more efficient method is to review all the for a given form, with the option to drill down into a patient's data more wholistically as needed. This also coincides with how data is presented in Excel, which is how a lot of medical monitors are used to ingesting data.

If we create a configuration file, this could be a feature that may turn on/off depending on the size of the study.

PR #146, opened 2024-11-21
Finally implement row review

### LINKED ISSUE #99: Allow review of singular records / rows

jthompson-arcus:
Closes #99


LDSamson:
@jthompson-arcus Maybe good to discuss the conceptual UI implementation together.

jthompson-arcus:
> @jthompson-arcus Maybe good to discuss the conceptual UI implementation together.

I am happy for all input on the UI implementation. There are several ways it could go and no clear best option.

jthompson-arcus:
@LDSamson & @aclark02-arcus this PR isn't quite done but I'm looking for some feedback. Just take it for a spin and see what you think from a UI perspective (look at the code if you want).

Still needed to be done (but running out of steam to complete):
- [ ] Indication of how many records are being reviewed (counts? progress bar?)
- [ ] Displaying record level comments.
- [ ] Handling of forms where review is not required.

jthompson-arcus:
@LDSamson & @aclark02-arcus just added a rudimentary progress bar.

aclark02-arcus:
@jthompson-arcus, At first when I saw your gif, I thought, "whaaaaa?" But now that I've pulled the code down and played with it, I think the UI components you added are really smart. The only improvement that I can think of would be to color code the highlighted rows the user changes with the same color in the progress bar. So, for example, below I've unchecked two rows. Instead of showing up as a light green/blue, maybe they should be a light red to match the progress bar?

image

jthompson-arcus:
I noticed that sorting or reordering table resets the checkbox visuals so putting the PR back on draft.

LDSamson:
Today I can't check more than shown below unfortunately due to technical issues (I am traveling by train, laptop is empty, power supply does not work). But I think the UI looks really nice already.
Here are some  thoughts regarding the UI (haven't properly looked at the code yet): 

- What if you would use the CSS theme colors instead (with the CSS labels success/warning/info): 
https://bootswatch.com/spacelab/
Then they will fit a bit better in the current theme, and they will change easily if you set a different theme.
- When selecting 'show all patients' the selection does not work as expected anymore on all rows. 
- For the longer run: I think it would feel natural if there is a checkbox next to the name of the review status column that will select everything in the table (thus similar to the save review check box, but closer to the data). 

Maybe we can just disable (or simply hide if that is easier) the column with selection boxes in cases that are not yet well implemented? (E.g. when using a table with data of all patients, or when showing a form of which review is not required).

In addition, could you write down the functionality that you added in a few short sentences so that we can use that as feature descriptions for our documentation? Can also be in a comment here initially.

Thanks for all the work! Again, it looks really nice already!

LDSamson:
> I noticed that sorting or reordering table resets the checkbox visuals so putting the PR back on draft.

I noticed it also resets these visuals if there is a lot of data in the table and you scroll down (probably related to the `Scroller` datatables plugin only rendering the part of the data that is visible)

jthompson-arcus:
The issue with the table display was quite extensive. The problem is that there are three data objects that need to be in sync: the DOM itself, the JS DataTable object, and the server data. To address this problem, I have divorced the datatable render from a reactive and used a proxy object for the table to update the data.

There are some downsides from a developer perspective in doing this:

- The form modules are now more complicated with more observers. One needs to take care about what events should be triggered from which actions.
- The active data is no longer directly tied to what is being viewed in the UI. This impacts tests.

R/fct_tables.R
```diff
@@ -51,12 +51,23 @@ create_table.default <- function(
...
+    data <- dplyr::mutate(
+      data,
+      o_reviewed = dplyr::case_when(
+        any(reviewed == "No") & any(reviewed == "Yes") ~ list(list(reviewed = NA, ids = id)),
+        any(reviewed == "Yes") ~ list(list(reviewed = TRUE, ids = id)),
+        .default = list(list(reviewed = FALSE, ids = id))
+      ),
+      .by = dplyr::all_of(keep_vars))
+    keep_vars <- c("o_reviewed", keep_vars)
+  }
```
LDSamson, line 64: Still looking into the structure of everything, including the `o_reviewed` object. But I feel like this is an important (first?) step in the entire process, right? What is reviewed now, is that only `TRUE` if all items are reviewed?

R/mod_common_forms.R
```diff
@@ -93,60 +93,236 @@ mod_common_forms_server <- function(
...
+              )),
+            rowCallback = row_callback
+          ))
+      })
+      SAE_proxy <- DT::dataTableProxy("SAE_table")
+    }
+    
+    output[["common_form_table"]] <- DT::renderDT({
+      datatable_custom(
+        isolate(subset(common_form_data(), input$show_all_data | subject_id == r$subject_id)),
```
LDSamson, line 179: `isolate` here is needed in order to prevent some refreshing loop/because the data is updated using the `dataTableProxy`, right?

R/mod_common_forms.R
```diff
@@ -93,60 +93,236 @@ mod_common_forms_server <- function(
...
+      DT::reloadData(common_form_proxy)
+      req(SAE_data())
+      DT::reloadData(SAE_proxy)
+    }, ignoreInit = TRUE)
+    
+    observeEvent(r$subject_id, {
+      req(common_form_data())
+      reload_data(reload_data() + 1)
+      df <- common_form_data() |> 
+        dplyr::mutate(o_reviewed = Map(\(x, y) modifyList(x, list(updated = NULL, disabled = y)), o_reviewed, subject_id != r$subject_id))
```
LDSamson, line 302: Does the `disabled` variable currently do anything? Or does it not yet have a function?

R/fct_tables.R
jthompson-arcus: Correct, it is only true if all records contained in the row are marked as `reviewed`. Naming could be better. It's kind of an "overall reviewed" for the row.

R/mod_common_forms.R
jthompson-arcus: Yes, `isolate()` causes the render to only run once.

R/mod_common_forms.R
jthompson-arcus: The `disabled` variable was added to handle when all subjects are being viewed. This value is used in the datatable object. The checkbox gets rendered as disabled if you aren't currently viewing that subject.

R/mod_review_forms.R
```diff
@@ -131,16 +172,14 @@ mod_review_forms_server <- function(
         # it will give a warning. This would be rare since it would mean a datapoint with the same edit date-time was reviewed but another one was not. 
         # probably better to use defensive coding here to ensure the app does not crash in that case. However we need to define which review status we need to select
         # in this case get the reviewed = "No"
-        review_status <- with(review_data_active(), reviewed[edit_date_time == max(as.POSIXct(edit_date_time))]) |> unique()
-        review_comment <- with(review_data_active(), comment[edit_date_time == max(as.POSIXct(edit_date_time))]) |> unique()
-        if(length(review_status) != 1) warning("multiple variables in review_status, namely: ", 
-                                               review_status, "Verify data.")
+        review_status <- unique(review_data_active()[["reviewed"]])
```
LDSamson, line 175: I think `review_status` can now contain both `TRUE` and `FALSE`, based on differences in edit_date_time. I this intended? Or is this not true?

R/app_server.R
```diff
@@ -30,6 +30,9 @@ app_server <- function(
     })
   check_appdata(app_data, meta)
   
+  session$userData$review_records <- reactiveValues()
+  session$userData$update_checkboxes <- reactiveValues()
```
LDSamson, line 34: If I understand it correctly, the variable 'update_checkboxes' is only `TRUE`, `FALSE`, or `NULL`, based on the value of `input$form_reviewed` . Why is it called `update_checkboxes` and not `form_reviewed` or something similar, for consistency?

R/mod_study_forms.R
```diff
@@ -193,6 +197,96 @@ mod_study_forms_server <- function(
       lapply(add_missing_columns(item_info, cols)[1, cols], isTRUE)
     })
     
+    observe({
+      reload_data(reload_data() + 1)
+      session$userData$update_checkboxes[[form]] <- NULL
+      session$userData$review_records[[form]] <- data.frame(id = integer(), reviewed = character())
+    }) |> 
+      bindEvent(r$subject_id, r$review_data,
+                ignoreInit = TRUE)
```
LDSamson, line 206: This feels a bit duplicated with the observeEvent in `mod_review_forms` and `mod_common_forms`. Placing a cat_dev in these three events, and updating subject_id fires multiple updates:
image

R/fct_tables.R
LDSamson: I wonder if we should extract this in a separate helper function since it is now such a pivotal step. That way, we can explain what is going on in a Roxygen tab, and we can write some simple tests for it. The o_reviewed object is now completely undocumented, it is hard to see what is needed in this object, what columns we expect here, and which would trigger an update etc.

R/mod_common_forms.R
```diff
@@ -93,60 +93,236 @@ mod_common_forms_server <- function(
...
+    # Ensure that review data gets reset whenever the review data or subject
+    # being viewed is changed.
+    observe({
+      reload_data(reload_data() + 1)
+      session$userData$update_checkboxes[[form]] <- NULL
+      session$userData$review_records[[form]] <- data.frame(id = integer(), reviewed = character())
+    }) |> 
+      bindEvent(r$subject_id, r$review_data,
+                ignoreInit = TRUE)
+
```
LDSamson, line 211: Just wondering: do you have a specific reason why you sometimes use observe/bindevent and sometimes `observeEvent`?

R/mod_common_forms.R
```diff
@@ -93,60 +93,236 @@ mod_common_forms_server <- function(
...
+            subset(r$review_data, subject_id == r$subject_id & item_group == form),
+            by = "id"
           ) |> 
-          dplyr::select(-dplyr::starts_with("SAE"))
+          # Only update records where the review status is being changed
+          dplyr::anti_join(
+            subset(r$review_data, subject_id == r$subject_id & item_group == form),
+            by = c("id", "reviewed")
+          ) |> 
+          dplyr::arrange(id)
```
LDSamson, line 255: the upsert/filter/semi_join and anti_join are a bit hard to follow for me. Is the semi_join really needed for example? Can't you just add subject_id and form to the filter directly?

Also, I think this part could be a candidate to extract in a helper function since it is duplicated in `mod_review_forms` and is more or less 'business logic'

LDSamson (commented):
I added a view comments. I will have to stop now, and and probably can only continue in the new year. In general, I like the way it is implemented, but I am just wondering if we can make it a bit simpler and easier to follow. It is a bit hard to follow what happens when exactly, in what order.

Have a nice Christmas and new year!

R/mod_review_forms.R
jthompson-arcus: It can contain both "Yes" and "No" and that is intentional. It is no longer attached to `edit_date_time`, but instead grabs the values directly from the `reviewed` field of the active review data. We no longer need `edit_date_time` because records are now unique in the database.

R/app_server.R
jthompson-arcus: Naming conventions are a bit in the eye of the beholder, but I named it `update_checkboxes` because that is the action the reactive is meant to trigger. Plus it would be inaccurate to call it `form_reviewed` because the value is more tied to changes in the sidebar checkbox, not to the review status of a form.

R/mod_study_forms.R
jthompson-arcus: Yes, each form is it's own module. I would expect it to fire for every module whenever the subject is changed or the review data is updated. I think this can be moved outside of those modules though and be contained in the review forms module.

R/mod_common_forms.R
jthompson-arcus: I prefer using `bindEvent()` when there are multiple triggers. In my opinion it is more readable.

R/mod_study_forms.R
```diff
@@ -193,6 +197,96 @@ mod_study_forms_server <- function(
...
+        dplyr::mutate(o_reviewed = dplyr::if_else(subject_id == r$subject_id, 
+                                                  lapply(o_reviewed, modifyList, list(updated = checked)),
+                                                  o_reviewed))
+      table_data(df)
+    })
+    
+    observeEvent(input$table_review_selection, {
+      # Update review values for session's user data
+      session$userData$update_checkboxes[[form]] <- NULL
+      session$userData$review_records[[form]] <-
```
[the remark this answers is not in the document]
jthompson-arcus: In essence it is still being used in that way because that's where the field `reviewed` is coming from. The new field `o_reviewed` is necessary for the datatable object. Each datatable is pivoted, so the datatable itself would contain multiple records, this needs to be consolidated in some way for a row level review to even be possible.

R/mod_common_forms.R
jthompson-arcus: I will take a look.

R/mod_study_forms.R
jthompson-arcus: On further inspection, I don't think we really want to change any of this observer. There doesn't appear to be any advantage to further abstracting this logic. It's important to ensure the datatable's server-side object is getting reloaded whenever the subject is changed of the review data is updated.

R/mod_common_forms.R
jthompson-arcus: I cannot add `subject_id` and `item_group` to the filter directly because `input[[review_selection]]` does not contain those values. It only contains the `id`'s. This was intentional redundancy. Even though `session$userData$review_records[[form]]` gets reset whenever `r$subject_id` is changed, this ensures that no unintended records are being assessed. The overhead is very little because this is only triggered when a checkbox in a datatable is clicked.

jthompson-arcus:
With all the requests for helper functions and such, I decided it was best to move all this logic to it's own module. Now all changes can be made in one location and the differences between the forms are essentially eliminated. The module itself still needs to be documented.

The PR grows and grows...

R/fct_tables.R
jthompson-arcus: https://github.com/openpharma/clinsight/blob/f5fa833a2b5e9cd6caed609b27ede955d87b34b2/R/fct_tables.R#L71-L97

jthompson-arcus: https://github.com/openpharma/clinsight/blob/f5fa833a2b5e9cd6caed609b27ede955d87b34b2/R/fct_form_helpers.R#L62-L88

R/mod_common_forms.R
jthompson-arcus: https://github.com/openpharma/clinsight/blob/f5fa833a2b5e9cd6caed609b27ede955d87b34b2/R/fct_form_helpers.R#L1-L39

R/app_server.R
LDSamson: The reason why it confuses me is because the value is a logical and that the meaning of the logical is whether the form is fully reviewed (or fully 'not reviewed'), not whether the checkboxes need to be updated. For example, I would not expect check boxes to update if `update_checkboxes` is `FALSE` for a specific form. But anyways, this is not a big deal, it works. It just took me some time to grasp what you are doing here.

R/mod_study_forms.R
```diff
@@ -162,30 +163,37 @@ mod_study_forms_server <- function(
...
+          dplyr::mutate(o_reviewed = Map(\(x, y, z) append(x, list(row_id = y, disabled = z)), 
+                                         o_reviewed, 
+                                         dplyr::row_number(),
+                                         subject_id != r$subject_id))
+      }
+      table_data(df)
     })
+    observeEvent(table_data(), {
+      mod_review_form_tbl_server("review_form_tbl", r, table_data, form, reactive(input$show_all), table_names)
+    }, once = TRUE)
```
LDSamson, line 196: Why is this wrapped in an observeEvent that is called once and not just call the module directly? Is it so that it is not run if table_data() is NULL? But shouldn't the table object always be created and be completely empty, even if no one has data in it?

R/mod_review_form_tbl.R
```diff
@@ -0,0 +1,172 @@
...
+                        outputId = table_proxy$rawId)
+    })
+    # Any time the review table is updated, "show all" is toggled, or the
+    # subject being viewed is changed, the datatable should be reloaded to show
+    # the new data
+    observeEvent(reload_data(), {
+      req(!is.null(show_all()))
+      req(table_data())
+      DT::reloadData(table_proxy)
+    }, ignoreInit = TRUE)
```
LDSamson, line 117: Is there a specific reason why DT::reload_data() is not in the observer above? I feel like these are always triggered together anyways.. 
But maybe I am missing something.
When `r$review_data` changes, `table_data()` will also be updated and thus the observer will also trigger.

jthompson-arcus: It is separate because we do not want to reload the data table when a user clicks on a checkbox. In that situation we only want to update the datatable object and the server data source. Otherwise, the table would reset which is not a good user experience.

jthompson-arcus:
> In my browser logs, it shows that the tables with the approriate table id's cannot be found. Maybe something is rendering too eagerly? image

It's kind of the opposite. It's because the datatables are not rendered yet.

R/mod_study_forms.R
jthompson-arcus: While what you said is true in the app, it's not true for all the tests. There is probably a better solution than having this observer.

R/mod_review_form_tbl.R
```diff
@@ -0,0 +1,175 @@
...
+    observeEvent(show_all(), {
+      req(table_data(), datatable_rendered())
+      reload_data(reload_data() + 1)
+      index <- match("subject_id", colnames(table_data())) - 1
+      if (show_all()) {
+        DT::showCols(table_proxy, index)
+      } else {
+        DT::hideCols(table_proxy, index)
+      }
+    })
```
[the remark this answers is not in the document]
jthompson-arcus: In what way is it not functional right now?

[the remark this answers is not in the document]
LDSamson: We can also do this in a follow-up PR if you think this is a good idea.

LDSamson:
Things to consider for a change, but can be a follow up PR:

- It would be great if the check boxes are hidden if a form is marked as not required to review. Maybe by passing the `forms_to_review_data` from the app_server to the `r` object
- think about splitting o_reviewed data from the table in shiny (and just combining them in the datatable object). I think this would help with the logic and make the reactive flow more natural (although I would have to try it out a bit to confirm this).

R/mod_review_form_tbl.R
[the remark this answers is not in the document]
LDSamson: With 'not functional' I meant the enhancement mentioned in #22 (reviewing data of all subjects at the same time in a form) is not yet functional.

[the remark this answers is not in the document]
jthompson-arcus: Got it.

I think we should leave the column visible because the checkboxes are disabled for all subjects not currently being reviewed. The current functionality is consistent with how reviews worked previously.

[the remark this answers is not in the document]
LDSamson: Hmm do you have a specific reason to leave the column visible? Are there specific benefits that I am missing? I will try to explain. For me personally, having the check boxes visible when viewing all subjects feels a bit awkward since viewing all subjects is about viewing general trends, not about reviewing one subject's data. Hiding the check boxes feels more natural in such a state. Furthermore, the disabled state is a bit hard to see, and it is not immediately clear at a first glance as to why they are disabled, which can cause confusion. 

@aclark02-arcus what do you think here?
Again, we can discuss later as well, it is not a deal breaker since it all works, it's more about optimization/user experience. I can also ask some feedback from users internally first.

LDSamson:
@aclark02-arcus do you have any comments on this PR?

R/mod_review_form_tbl.R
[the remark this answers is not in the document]
aclark02-arcus: I do see @LDSamson's point. Users probably wouldn't consider reviewing the selected patient while viewing all patients. So from that perspective, I see why hiding the `Review Status` column would cause less confusion. However, I could go either way with this one.

One thing that is WAY more confusing (to me) when expanding to show all patients is that all patient data (1) doesn't show the existing patient first and (2) the rows aren't sorted by patient.  Everything is just kind of "jumbled". Sorry, I know this isn't really related to the topic at hand. 🤷🏼‍♂️ 

image

R/app_server.R
aclark02-arcus: naming things is hard.

aclark02-arcus:
> @aclark02-arcus do you have any comments on this PR?

Yes, it's beautiful! @LDSamson, we've been using this internally now for a while... well, a version of this PR, before its the review function was modularized which is a really nice touch. Huge improvement! It has my blessing.

LDSamson (approved):
Thanks for adding the comments in the javascript section. 

I think we should proceed with merging this PR. There are some things that can be optimized (see all the comments), but that can be done in future PRs.

One thing that is in my opinion most important after merging this, is creating new unit tests/ end to end tests to capture the new behavior. All major features should be tested (ideally using the same BDD test descriptions as used before) before we can use the version in production.

PR #149, opened 2024-12-05
Fix `dblclick_to_form()`

jthompson-arcus:
It was discovered that `dblclick_to_form` did not work correctly when a table was sorted. This PR seeks to fix that bug.


jthompson-arcus:
No NEWS item was added because the double-click feature was added in the current development iteration.

LDSamson (approved):
Looks good, thanks. Only question is this: is it possible to add a custom JS click on one of the columns in the app-feature-1 test so that this bug is covered? just before the code shown below. I think it would be a nice addition. For the rest no other comments

https://github.com/openpharma/clinsight/blob/6954ed42907e082629d9e8ac5c6789ca00c1ad7c/tests/testthat/test-app_feature_01.R#L44-L45

EDIT: maybe something like this?
```
app$run_js('$("#start_page_1-overview_table th").filter(function() {return $(this).text() == "Status"}).closest("th").trigger("click")')
```

LDSamson (approved):
LGTM!

PR #150, opened 2024-12-13
Fix `pageLength` bug

jthompson-arcus:
The DataTable extension `Scroller` is disabled on a couple query tables, but the `pageLength` is still set at 10. This PR changes the `pageLength` to all rows. Typically there is not a lot of data in these tables, so enabling all rows shouldn't be a problem.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: clinsight
 Title: ClinSight
-Version: 0.1.1.9011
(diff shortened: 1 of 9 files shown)


aclark02-arcus (approved):
LGTM Jeff. Thank you!

LDSamson (approved):
Looks good to me. Thanks, I did not yet notice this issue!

PR #151, opened 2024-12-19
Improve event labels

### LINKED ISSUE #67: Need to Generalize `add_timevars_to_data()`
https://github.com/openpharma/clinsight/blob/0eec2fd3e787a27733d9ad28d5d5609e9247c2c3/R/fct_data_helpers.R#L116-L159

We need to find some way to generalize this function and remove all the hard coding. I don't know if this means we need another sheet/item in the metadata or not.

LDSamson:
Closes #67. This PR adds a lot of flexibility in creating event/visit labels.

The event tab in the metadata now looks like this: image

The new setup enables the user to set the logic on which the event labels are created, and to set the names (based on prefix and suffix). 

In addition, if a variable with the name 'event_name_edc' is available in the data, and no suffix is set in the table, the event_name_edc will be added to the visit names:
image
in the timeline: 
image

Lastly, the labels in the compact timeline top right are now the number of 'expected events' but this expand flexibly if more visits occur. 
image
The only thin less flexible are the labels for this compact timeline; these are now always V0-Vx. I don't know how you guys think of this.

The metadata needs to be updated to be able to benefit from this update.


jthompson-arcus:
@LDSamson I will give it a look. First impressions is that this PR actually reduces flexibility for us. We built the events tab off of the visits, so instead of using V0-Vx (or Visit 1 - Visit X), we were using C1D1-CXDY and there applicable names. This is also why we wanted the additional flexibility during `merge_meta_with_data()`. Further hard-coding the V0-Vx notation in the app wouldn't be a direction we would want to go. Our events tab looks like this:

| event_number | event_name | event_label |
| --- | --- | --- |
| 1 | Screening | SCR |
| 2 | Cycle 1 Day 1 | C1D1 |
| 6 | Cycle 1 Day 3 | C1D3 |
| 20 | Cycle 2 Day 1 | C2D1|

LDSamson:
I was not aware that you customized the events tab. This was previously only used for the compact timeline. On a side note, are you using the event_number for something specific? I don't think that was used in clinsight.

'Hard coding' was done for the small timeline but this can be changed. I think the rest is much more flexible in this way, also the normal 'Visit ...' notation is now completely customizable. You can create the same labels as you have now by filling in the table like this, if these patterns match the `event_id` in the raw data:

image

But you can still provide your own `event_name` (or other variables) if you prefer to do so, that did not change. 

One advantage of this PR is that there can be many more visits, for example during follow-up. The number of visits do not not necessarily have to be hard coded in advance. And it provides a way to deal with unscheduled visits.

LDSamson:
Marking as draft until custom visit labels can be used in the compact timeline again

LDSamson:
Okay, I made improvements and restored the option to customize the compact timeline labels. @aclark02-arcus can you have a look if it works for you? 

The events tab in the metadata now looks like this:

grafik

Green columns are optional. From the orange ones, they can be used both but at least one needs to be available.
You can now just fill in the event_id column (previously `event_label`) and nothing else, and everything should be as expected.

Few more details; 
- Only visits of which `is_expected_visit` is `TRUE` will show up in the timeline. Especially useful for 'unscheduled visits' which can happen at any time and should not show up here.
- Events will be created based on matching with event_id in the study data. 
- The `event_id` will be taken as label for the compact timeline, but this can be overwritten by `event_label_custom`
- If `event_id_pattern` is used to match events, multiple events can match, and the labels will default to `V0-V1-...-Vx` for these events. 
- The longer `event_name` will also default to `event_id` but can also be customized using `event_name_custom`. A sequence number will be added to the custom name if needed (`add_sequence_to_name`). If the column `event_name_edc` is available in the data, this value will be added as a suffix in brackets:
grafik


Can you check if it works for you? 
We need this construction mostly because we have studies with an indeterminate number of 'normal' visits and thus need some flexibility in creating event labels. I think this generalizes it to

aclark02-arcus (changes requested):
Hi @LDSamson, this PR had a couple issues for our data sources. Specifically, we weren't the `static_overview_data` on the Start tab wasn't showing all the `event_id`s in the Visit column. In addition, the `fig_timeline()` wasn't plotting any visits for most patients.

Recapping our conversation today, it looks like merging `events_table` on `vis_num` seems to be the primary issue when attempting to apply this PR's code to our data. As shared, we have certain `event_id`s that have various `vis_num` values (including `SCR` values), but the `event_table` data only stores [the max `vis_num`](https://github.com/openpharma/clinsight/blob/cf8ddd92a7adb909e55d0e38c78abcd007a2a08b/R/fct_data_helpers.R#L259). Hence, when you join on `event_id` AND `vis_num` [here](https://github.com/openpharma/clinsight/blob/cf8ddd92a7adb909e55d0e38c78abcd007a2a08b/R/fct_data_helpers.R#L282). When I just merge on `event_id`, however, the code appears to work as intended.

LDSamson:
@aclark02-arcus thanks for checking. I think that some subjects in your data have multiple screening days, is that correct? If so, that might have been the root cause of this issue.

R/fct_data_helpers.R
```diff
@@ -296,8 +243,6 @@ get_meta_vars <- function(data = appdata, meta = metadata){
   if(length(data) == 0) stop("Empty list with data provided")
   vars <- list()
   # add metadata variables:
-  vars$events <- setNames(meta$events$event_label, meta$events$event_name)
-  
   vars$items <- meta$items_expanded |> 
     dplyr::distinct(item_name, item_group) |> 
     split(~item_group) |>
```
aclark02-arcus, line 248: Okay, so after investing a good amount of time trying to merge `dev` with our internal version of `dev` so that I can isolate just the changes from this branch internally, it's become apparent that that the two branches have diverged too much. That is, it's not easy to compare new additions in a timely manner... mostly because of how we use `vars$items` internally. I don't know when it was changed on the open-source repo, but this new manner of grabbing `item_names` by `item_group` is actually pretty similar to what I implemented internally, but different. Sooo, its not easy to quickly compare. Sorry @LDSamson, I blame myself for not getting my PR that creates listings by `form_id` submitted faster, but that's the culprit.

aclark02-arcus: I will try another method to test out our data on this PR, but I just wanted to make the note above because "its time" that I get that PR submitted... otherwise our repos diff is too messy to untangle.

aclark02-arcus (approved):
Edit: whoops - PR approved by mistake. Please disregard.

R/fct_figures.R
```diff
@@ -84,33 +84,23 @@ fig_boxplots <- function(
...
-    dplyr::mutate(
-      event_label = factor(event_label, levels = events$event_label)
-    )
-  all_events <- events |> 
-    dplyr::mutate(event_name = factor(event_name, levels = event_name), 
-                  event_label = factor(event_label, levels = event_label))
-  uneven_events   <- all_events[1:length(all_events$event_name) %% 2 == 0, ]
-  even_events     <- all_events[1:length(all_events$event_name) %% 2 != 0, ]
+  labels_in_data <- unique(na.omit(data$event_label))
+  all_labels <- levels(data$event_label)
```
aclark02-arcus, line 97: This is returning `NULL` for us at the moment. There is data populating this column, its just not a factor. IE, `is.factor(data$event_label) returns `FALSE`, and `unique(data$event_label)` can't be used because the factor levels would help retain visits that we're not completed in the timeline. I'll drill down and figure out why.

R/fct_data_customize_events.R
```diff
@@ -0,0 +1,253 @@
...
+      derived_event_order <- derived_event_order |> 
+        # selects the most unique order per event_id:
+        dplyr::mutate(order_occurrence = dplyr::n(), .by = derived_order) |> 
+        dplyr::slice_min(order_occurrence, by = event_id, with_ties = FALSE) |> 
+        dplyr::select(-order_occurrence) |> 
+        dplyr::arrange(event_id, derived_order) |> 
+        dplyr::mutate(derived_order = dplyr::row_number())
+    }
+  }
+  events_table <- events |>
```
LDSamson, line 193: @aclark02-arcus does this internal table show the events correctly in your own study data if you check it with a browser statement, in particular the `event_label_custom` variable?

If not, could it be that you are changing the event_label in a custom function later on during the merge process? with one of these pre_pivot_fns | post_pivot_fns | post_merge_fns ?

LDSamson: @aclark02-arcus another thing I was wondering: are you testing this PR branch directly or are you testing it together with other changes that you made in your local branch? If the latter is the case: Is the issue you mentioned actually because of this PR or due to changes that you made in your local version?

aclark02-arcus: Nope, we do have a custom event_label `mutate()` but I turned that off for my testing of this branch.

R/fct_figures.R
aclark02-arcus: Turns out it was just a simple mistake I made during the deployment process. Sorry for raising a flag pre-maturely.

PR #152, opened 2024-12-20
Improve timeline

LDSamson:
See below for the timeline improvements:

timeline improvements

The PR works but it needs some fine tuning. To optimize the improvements, I need to add a custom `js` function to trigger the `redraw()` option for the timeline when hovering over an item that has been reviewed, and after clicking on such an item, to get a proper animation in such a case. But I was not yet able to find the right tag for these timeline boxes. Would you able to help with this, @jthompson-arcus ?


LDSamson:
@jthompson-arcus I've found the issue with the class id and the javascript functions: it cant find the tag because the HTML widget renders late. https://stackoverflow.com/a/62796035/11856430.

I will add a few more tweaks.

LDSamson:
I added a custom hook for redrawing the timeline on item selection. It looks now like this: 

clinsight_timeline_improvement

Other customizations:

- The treatment label can be customized using the metadata `treatment_label` setting:
grafik
- The on-hover labels are now more consistent and always show the dates first
- The treatment on-hover label also show the treatment dose (if available)

R/fct_appdata_summary_tables.R
```diff
@@ -76,34 +91,60 @@ get_timeline_data <- function(
...
   drug_data <- if(is.null(data$General)){
     data.frame()
   } else{
-    data$General |> 
-      dplyr::filter(item_name %in% c("DrugAdminDate", "DrugDiscontDate")) |> 
+    df_drug_admin <- data$General[
+      data$General$item_name %in% c("DrugAdminDate", "DrugAdminDose"), 
+    ] |> 
+      tidyr::pivot_wider(names_from = item_name, values_from = item_value) |> 
+      clinsight::add_missing_columns(c("DrugAdminDate", "DrugAdminDose")) |>
```
jthompson-arcus, line 111: Why `clinsight::`?

jthompson-arcus (changes requested):
@LDSamson I like it a lot! In addition to the comments below, I noticed that `metadata.rda` wasn't added. So `clinsight::metadata` doesn't show the treatment label.

R/app_server.R
```diff
@@ -207,10 +207,13 @@ app_server <- function(
       select = (i == common_forms[1])
     )
   })
+  treatment_label <- meta$settings$treatment_label %||% "\U1F48A T\U2093"
```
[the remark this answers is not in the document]
LDSamson: I moved it to `get_metadata()` and documented it. I also impute it now in `mod_timeline_server()` so that it will not crash on older metadata files where treatment label is missing (a bit redundant indeed, but just to be sure).

R/fct_appdata_summary_tables.R
LDSamson: oops

LDSamson:
@jthompson-arcus I also added a few tests

jthompson-arcus (approved):
LGTM

PR #153, opened 2024-12-30
Add "sophisticated" excel download button to `dev`

### LINKED ISSUE #96: Allow for excel / csv export of listings

aclark02-arcus:
Officially closes #96


aclark02-arcus:
Hi @LDSamson, looks like I was able to use my old branch that merged into `devex` without many conflicts! I tested, and it works as expected.

R/fct_data_helpers.R
```diff
@@ -505,6 +505,35 @@ add_missing_columns <- function(
...
+        filename = paste("clinsight", gsub(" ", "-", table_name), sep = ".")
+        )))
+  } else {
+    dt_dom <- default_args$dom |> eval()
+    dt_exts <- default_args$extensions |> eval()
+    dt_opts <- default_args$options |> eval()
+  }
+  return(list(dom = dt_dom, exts = dt_exts, opts = dt_opts))
+}
+
```
LDSamson, line 536: Since most tables will now receive this download button (if set in configuration), and thus being more or less the 'standard': I feel like this can be simplified by including this logic in `datatable_custom`. We only have to add a `(download)_table_name` argument, and an argument to override the `allow_listing_download` button for tables that are not supposed to get a download button. It prevents the need for extracting `datatable_custom` arguments, editing them, and then putting them back in `datatable_custom`. 

@aclark02-arcus would you mind if I push an example of that here? If you don't like it we can revert it.

R/fct_data_helpers.R
```diff
@@ -505,6 +505,35 @@ add_missing_columns <- function(
...
+#' @keywords internal
+#' @return list with three named objects: `dom`, `exts`, and `opts`
+dt_config <- function(data, table_name = "form") {
+  default_args<- formals(datatable_custom)
+  if(nrow(data) > 0 & isTRUE(get_golem_config("allow_listing_download"))) {
+    dt_dom <- 'Bfti'
+    dt_exts <- c("Buttons", eval(default_args$extensions))
+    dt_opts <- list(buttons=list(list(extend = 'excel',
+        text = '<i class="fa-solid fa-download"></i>',
+        filename = paste("clinsight", gsub(" ", "-", table_name), sep = ".")
```
LDSamson, line 527: Only other comment/suggestion I have is to add a title argument in here, so that the title inside the Excel download table is prettier (my suggestion would be something like the table_name with 'extracted from ClinSight')

LDSamson (commented):
Looks good and works great. I just have two suggestions for simplification/improvement, see comments.

R/fct_data_helpers.R
aclark02-arcus: I'm not opposed to this at all. Jeff asked me the same question a long time ago and I think I opted for this approach since not all tables get the download, and I didn't want make custom_datatable more complicated with extra arguments... I think...

If you want to push an example in another branch so I can compare, that'd be great! But you don't have to if you don't want to. Completely up to you @LDSamson!

LDSamson: Oops, accidentally pushed the suggestions to both this branch and a different branch. Restored it now

aclark02-arcus: Reviewed #157. I'll resolve this conversation.

aclark02-arcus:
marking as draft until #157 get's merged.

aclark02-arcus:
Great adds @LDSamson. I would merge this one myself, but I don't want to do so prematurely in case you wanted to check anything first?

LDSamson (approved):
Looks good to me! I've added a few minor tweaks, and a very basic unit test.

PR #154, opened 2024-12-30
Switch legend to say "significance pending"

aclark02-arcus:
... Instead of "Significance Unknown". This was discussed in a previous dev meetup, but essentially, the word "unknown" implies the chance that we may never know, but in reality, DM teams are usually working to correct those omissions in the EDC and will eventually be corrected. Thus, "pending" is a slightly better descriptor.

image


data-raw/internal_data.R
```diff
@@ -5,9 +5,10 @@
 col_palette <- c("#1a9641", "#a6d96a", "#fdae61", "#d7191c", "grey50")
 names(col_palette) <- c("within limits", 
                         "out of limits, clinically insignificant", 
-                        "out of limits, significance unknown", 
+                        "out of limits, significance pending", 
                         "out of limits, clinically significant", 
                         "limits unknown")
+usethis::use_data(col_palette, overwrite = TRUE)
```
aclark02-arcus, line 11: Note: this line was needed to update `data/col_palette.rda`. I'm not sure if it's used anymore, but I wanted to make sure it got updated too. `col_palette` is still present in `sysdata.rda` too.

LDSamson (commented):
Maybe we should also delete the `data/col_palette.rda` in this PR to avoid confusion. For the rest it looks good to me

data-raw/internal_data.R
aclark02-arcus: deleted!

aclark02-arcus:
@LDSamson, recent commit removes  `col_palette.rda` & `col_palette.Rd`. Should be ready now!

LDSamson (approved):

PR #156, opened 2024-12-30
Make query handling optional

### LINKED ISSUE #101: Config option to hide `Create Query` button
Our pilot users have made it abundantly clear that they will not be opening queries via the `clinsight` app since they are required to do it through our EDC platform, and the two don't talk to each other. However, we don't want to hide all the query functionality... our data management dept generates a "query report" weekly that (I haven't seen yet, but) we'd like to ingest into the database, so users know if there is an open query tagged to a certain record.

jthompson-arcus:
Closes #101


LDSamson (approved):
Looks good to me.

Should be documented but can be done in one go when we update the developer's manual in a future PR

PR #157, opened 2025-01-10
Table download button - suggestion for change

LDSamson:
As discussed here: https://github.com/openpharma/clinsight/pull/153#discussion_r1909039329


R/fct_data_helpers.R
```diff
@@ -576,6 +554,8 @@ datatable_custom <- function(
     plugins = "scrollResize",
     dom = "fti",
     options = list(),
+    allow_listing_download = NULL,
+    table_name = NULL,
```
aclark02-arcus, line 558: I would suggest changing the name of the `table_name` argument to avoid confusion with the usual `rename_vars = table_names` call for `datatable_custom()`. Potential options could be "listing_name" or "table_label" for example? Ideally, it would be more concise to change the metadata variable `table_names` to `table_variables` or `table_var_names`, but that may be too much effort for this PR.

R/mod_common_forms.R
```diff
@@ -154,13 +153,18 @@ mod_common_forms_server <- function(
...
+        rename_vars = table_names, 
+        rownames= FALSE, 
+        title = form, 
+        escape = FALSE, 
+        table_name = paste(
+          simplify_string(form), 
+          ifelse(input$show_all_data, "all_patients", r$subject_id), 
+          sep = "."
+        )
+      )
```
aclark02-arcus, line 167: Here is one of many examples where `rename_vars = table_names` and `table_name = paste(...)` shows up and I'd argue it's slightly confusing. As suggested above, perhaps switching to `table_label` would be more readable / discerning? Just a suggestion.

aclark02-arcus (commented):
Just a comment below for you to consider. Overall, this task was is executed much more concisely inside `datatable_custom()`. Looks great!

R/fct_data_helpers.R
LDSamson: I agree it's confusing. How about `export_label` since it's only used for exported tables?

aclark02-arcus (approved):
Looks great!

PR #160, opened 2025-01-24
Improve row review a little

jthompson-arcus:
Utilize lazy evaluation of `reactive()`


jthompson-arcus:
@LDSamson I know this isn't the tests and such that we needed to add. We did notice a delay on our deployment when saving a review. This PR takes advantage of the lazy reactivity we had before. Now when the data needs to be recalculated (change subject, form data, or a saved review), it will only get recalculated when the datatable needs to be rendered.

R/mod_study_forms.R
```diff
@@ -163,35 +163,36 @@ mod_study_forms_server <- function(
...
+        create_table(expected_columns = names(form_items)) |> 
+        dplyr::mutate(o_reviewed = Map(\(x, y, z) append(x, list(
+          row_id = y, 
+          disabled = z,
+          updated = isolate(session$userData$update_checkboxes[[form]]))
+        ), 
+        o_reviewed, 
+        dplyr::row_number(),
+        subject_id != r$subject_id))
     })
```
LDSamson, line 194: More a question to myself probably: would it work to use bindEvent here, so that it only updates when review data of that specific form is updated? ==> I will check if something like this is useful: 

```
|> 
  bindEvent(r$filtered_data[[form]], r$review_data[item_group == form, ], r$subject_id)
```

jthompson-arcus: I had thought about this but never implemented it. I'm not really a fan that this adds observers that trigger for every form twice. I think I would rather just have the table re-render whenever the review data changes even though it could cause a user to lose some work.

R/mod_common_forms.R
```diff
@@ -97,56 +97,83 @@ mod_common_forms_server <- function(
   moduleServer( id, function(input, output, session){
     ns <- session$ns
 
-    common_form_data <- reactiveVal()
-    SAE_data <- reactiveVal()
-    observe({
-      df <- {
+    common_form_data <- reactive({
+      cat(form, "data computed \n")
```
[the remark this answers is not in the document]
jthompson-arcus: Merged your edits that added `golem::cat_dev()` statements.

R/mod_study_forms.R
jthompson-arcus: I guess I also feel like if we are adding double `reactiveVal()` for the review data, why not restructure `r$review_data` to to be a reactive value list on forms like `r$filtered_data`?

LDSamson: > I had thought about this but never implemented it. I'm not really a fan that this adds observers that trigger for every form twice. I think I would rather just have the table re-render whenever the review data changes even though it could cause a user to lose some work.

Are you now talking about the change in mod_review_form_tbl that you did not like? Or all the changes that I suggested and that you now merged? 

I don't think it is adding more reactivity than there already was. It does the same as it did before, but less: it stops the reactivity as soon as the form-level review data is found to be unchanged. 
Alternatively, we can parse the rev_data_form() object to mod_review_form_tbl(), then we prevent duplication. 
Or place the entire data object calculation from mod_review_form to mod_review_form_tbl. 
Or, indeed, change it on the level of r$review_data to make it a list of reactive values so that each triggers on its own which would be a natural place to change it. But that will require changes in more modules and tests, that is why I did not yet suggest it 😄

jthompson-arcus: I know I merged in the `rev_data_form()` but I'm going to roll the back for this PR. I think the solution is to restructure `r$review_data` to be `r$review_data[[form]]`, then we can just use `r$review_data[[form]]` directly within `reactive()` and don't need to artificially add binding events, additional reactive values and observers.

R/mod_common_forms.R
```diff
@@ -97,58 +97,88 @@ mod_common_forms_server <- function(
...
+              reviewed == "No",
+              paste0("<b>", htmltools::htmlEscape(item_value), "*</b>"),
               htmltools::htmlEscape(item_value)
             )
+          ) |>
+          create_table(expected_columns = names(form_items)) |>
+          dplyr::mutate(o_reviewed = Map(\(x, y, z) append(x, list(
+            row_id = y, 
+            disabled = z,
+            updated = session$userData$update_checkboxes[[form]])
```
LDSamson, line 163: @jthompson-arcus before I forget: you might need to add back `isolate` when `bindEvent` is removed again

LDSamson (approved):

PR #161, opened 2025-01-28
Fix shinytest2 for Chrome browsers in which the old headless mode is deprecated (v132 and higher)

LDSamson:
Fixes issue described here: https://github.com/rstudio/chromote/issues/187 . Note that if we update the snapshot to 26Jan2025 or later, this additional line can be removed again since the setting is standard in Chromote v0.4.0 that just got released. 

https://github.com/rstudio/chromote/releases/tag/v0.4.0

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: clinsight
 Title: ClinSight
-Version: 0.1.1.9016
+Version: 0.1.1.9017
 Authors@R: c(
     person("Leonard Daniël", "Samson", , "lsamson@gcp-service.com", role = c("cre", "aut"),
            comment = c(ORCID = "0000-0002-6252-7639")),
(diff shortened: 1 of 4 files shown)


jthompson-arcus:
@LDSamson only question I have is do we want to add the option to individual test files so that `devtools::test_active_file()` will work without edits?

LDSamson:
@jthompson-arcus is it not working interactively? Locally it does for me but maybe I did somethign wrong. I placed the options in the shinytest2 set file instead of the testthat file.

jthompson-arcus:
> @jthompson-arcus is it not working interactively? Locally it does for me but maybe I did somethign wrong. I placed the options in the shinytest2 set file instead of the testthat file.

You know what, I just assumed it was in `testthat.R`. Doesn't reflect well on me as a reviewer lol. I'm not familiar with setup files. Learning something new today.

jthompson-arcus (approved):
🤌

PR #162, opened 2025-01-31
Create raw clinsightful_data

### LINKED ISSUE #105: Create raw CSV files for `clinsightful_data` object
Raw datasets need to be created for which the `clinsightful_data` object is built from. This will aid when changes are made to the metadata or the merging process.

LDSamson:
Closes #105


LDSamson:
All right guys, it would be great if you both can have a look here! This is a big change since we are changing the internal dataset, but after this we can always recreate it witthe the newly available raw data that I reverse engineered. Please have a look at dev/create_raw_clinsight_data.R and maybe try to run it with old clinsightful_data (the dataset currently in the dev folder). There are some changes made in the dataset, I tried to described them in the aformentioned R file,

tests/testthat/test-mod_study_forms.R
```diff
@@ -233,7 +239,7 @@ describe(
         app$expect_values(output = TRUE, export = TRUE)
         df <- app$get_value(export = "test-fig_data")
         expect_equal(as.character(unique(df$item_name)), "Temperature")
-        expect_equal(with(df, reviewed[subject_id == "NLD_06_755"]), c("No", "Yes", "No"))
+        expect_equal(with(df, reviewed[subject_id == "NLD_06_755"]), c("Yes", "Yes", "No"))
```
jthompson-arcus, line 242: @LDSamson why did this need to change?

LDSamson: Because the review status is set by random sampling above. The seed is still the same but the sorting within `appdata` changed and thus the assignment of review status changed as well.

jthompson-arcus (approved):
Everything looked good to me.

PR #166, opened 2025-02-13
remove even more unneeded reactivity in row review

LDSamson:
@jthompson-arcus is this something?

Note that this now also keeps userData of form x if only userData was saved from form y, which can be nice I think

### DIFF
--- a/R/mod_common_forms.R
+++ b/R/mod_common_forms.R
@@ -97,8 +97,16 @@ mod_common_forms_server <- function(
   moduleServer( id, function(input, output, session){
     ns <- session$ns
 
+    rev_data_form <- reactiveVal() 
+    observeEvent(r$review_data, {
+      rev_data_form_new <- with(r$review_data, r$review_data[item_group == form, ])
+      if(is.null(rev_data_form()) || !identical(rev_data_form(), rev_data_form_new)){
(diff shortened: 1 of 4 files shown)


R/mod_common_forms.R
```diff
@@ -172,11 +183,13 @@ mod_common_forms_server <- function(
...
-    
-    mod_review_form_tbl_server("review_form_tbl", r, common_form_data, form, reactive(input$show_all_data), table_names, form)
-    if (form == "Adverse events")
+      }) |> 
+        bindEvent(r$filtered_data[[form]], rev_data_form(), r$subject_id)
+      
       mod_review_form_tbl_server("review_form_SAE_tbl", r, SAE_data, form, reactive(input$show_all_data), table_names, "Serious Adverse Events")
+    }
+     
+    mod_review_form_tbl_server("review_form_tbl", r, common_form_data, form, reactive(input$show_all_data), table_names, form)
```
jthompson-arcus, line 192: I don't think it's a value add to put the SAE table module server into the `if` statement that creates the reactive. I think keeping both table module servers next to each other makes the code more readable.

jthompson-arcus (approved):
@LDSamson there are a couple changes I want to make, but since my branch is the base, I'm just going to merge and we can discuss it there.

PR #168, opened 2025-02-20
Improve form module reactivity

LDSamson:
As discussed, this PR will move all the logic for creating a (review) table from `mod_common_forms` and `mod_study_forms` to the already existing `mod_review_forms_tbl` module. In addition, only data from one form will be passed to the the modules.

This removes code duplication in the modules, it reduces reactivity (e.g. it will not reload all the table/figure data of form B when a review in form A is saved). 

In addition, I refactored some code in `mod_review_forms_tbl`, and fixed some smaller bugs in the `create_table` functions, such as data showing up as `NA` in the tables when data is new (because html bold tags are added in the app):
image
but showing as intended when data is reviewed:
image

Multiple tests are added to test the business logic for merging a table created with `create_table`.
Furthermore, a start was made to create a module test for the `mod_review_form_tbl` module.


R/mod_review_form_tbl_fct_helpers.R
```diff
@@ -1,3 +1,108 @@
...
+#' @keywords internal
+#'
+get_form_table <- function(
+    form_data,
+    form_review_data,
+    form,
+    form_items,
+    active_subject,
+    is_reviewed = NULL,
+    is_SAE = NULL,
```
jthompson-arcus, line 24:
```suggestion
    is_reviewed = FALSE,
    is_SAE = FALSE,
```
Not sure why we would set the defaults to NULL when the value is supposed to be a logical.

LDSamson: That is probably better indeed. Note that `is_reviewed` can be `NULL`/indetermined, but that is indeed not a reason to set the default at `NULL`, and not a reason for the default of `is_SAE` to be `NULL`.

jthompson-arcus: Maybe it makes more sense for `is_reviewed = NULL` since that's tracking the checkbox update item.

R/mod_review_form_tbl_fct_helpers.R
```diff
@@ -1,3 +1,108 @@
...
+#'
+#' @keywords internal
+#'
+get_form_table <- function(
+    form_data,
+    form_review_data,
+    form,
+    form_items,
+    active_subject,
+    is_reviewed = NULL,
```
jthompson-arcus, line 23: This argument feels poorly named because it is not the current review status but the overall pending review status. Having said that I can't come up with a better name, especially since it takes a logical. Maybe we should come up with a better name for both this argument and `update_checkboxes`.

R/mod_review_form_tbl_fct_helpers.R
```diff
@@ -1,3 +1,108 @@
...
+      o_reviewed = Map(
+        \(x, y, z) append(x, list(
+          row_id = y, 
+          disabled = z,
+          updated = is_reviewed)
+        ), 
+        o_reviewed, 
+        dplyr::row_number(),
+        if (is.null(active_subject)) FALSE else subject_id != active_subject
+      )
```
jthompson-arcus, line 66:
```suggestion
      o_reviewed = Map(
        \(x, y, z) append(x, list(
          row_id = y, 
          disabled = z,
          updated = if (!z & !is.null(active_subject)) is_reviewed)
        ), 
        o_reviewed, 
        dplyr::row_number(),
        if (is.null(active_subject)) FALSE else subject_id != active_subject
      )
```
I know you didn't introduce this issue but we only really want to be setting `updated` to TRUE/FALSE if `subject_id == active_subject`. I think in the application this would always just be NULL because whenever we initialize the table data we reset the overall pending review status. But better safe than sorry.

R/mod_review_form_tbl.R
```diff
@@ -59,10 +58,26 @@ mod_review_form_tbl_server <- function(
...
+        paste0("Warning: no data found in database for the form '", form, "'")
+      ))
+      golem::cat_dev(form, "| Computing merged data\n")
+      get_form_table(
+        form_data(),
+        form_review_data(), 
+        form = form, 
+        form_items = form_items,
+        active_subject = active_subject(),
+        is_reviewed = session$userData$update_checkboxes[[form]],
```
LDSamson, line 73: @jthompson-arcus Not much time to verify this today anymore, but shouldn't we replace 
        `session$userData$update_checkboxes[[form]]` with just `NULL` here?

LDSamson: Changed with 0fc8db5cc4837160d598fd2519a9d68fcfe500e2

R/mod_review_form_tbl_fct_helpers.R
LDSamson: Can we maybe even remove this `updated` logic, hard code NULL, and remove the `is_reviewed` argument? In this specific function

R/mod_review_form_tbl_fct_helpers.R
LDSamson: Should we also rename `updated` within `o_reviewed` to `pending_form_review_status` for consistency?

R/mod_review_form_tbl_fct_helpers.R
LDSamson: If that would cause problems Im fine with your suggestion

PR #169, opened 2025-02-24
Make `r$review_data` in a reactive values list

jthompson-arcus:


jthompson-arcus:
@LDSamson do you want to incorporate this PR into #168? I think it's a pretty lightweight solution to breaking up `review_data`.

R/mod_review_forms.R
```diff
@@ -112,9 +112,11 @@ mod_review_forms_server <- function(
...
     
     review_data_active <- reactive({
-      with(r$review_data, r$review_data[
-        subject_id == r$subject_id & item_group == active_form(),
-        ])
+      tryCatch(
+        subset(r$review_data[[active_form()]], subject_id == r$subject_id),
+        # Returns expected empty data frame for empty form
+        error = \(e) r$review_data[[names(r$review_data)[1]]][0,]
+      )
```
LDSamson, line 119: Since all forms will now always be available: is this now mostly needed for testing purposes?

R/mod_timeline.R
```diff
@@ -51,9 +51,9 @@ mod_timeline_server <- function(id, r, form, treatment_label = "\U1F48A T\U2093"
       req(timeline_data())
       # join below is on both form_repeat and item_group. Join by item_group 
       # is to prepare for future timelines when other forms will also be included. 
-      review_active <- with(
-        r$review_data, 
-        r$review_data[item_group == form & subject_id == r$subject_id, ]
+      review_active <- subset(
+        r$review_data[[form]],
+        subject_id == force(r$subject_id)
```
LDSamson, line 56: why is this `force` needed here? I don't know the function. What happens if it is not used?

jthompson-arcus: Oops. Meant to remove that.

tests/testthat/_snaps/mod_review_forms.md
```diff
@@ -1,42 +1,42 @@
...
+                                     item_name event_date      edit_date_time
+      Adverse events.1     Atrial Fibrillation 2023-08-15 2023-09-30 01:01:01
+      Adverse events.2                Cystitis 2023-09-01 2023-10-01 01:01:01
+      Vital signs.3    Systolic blood pressure 2023-07-01 2023-08-30 01:01:01
+      Vital signs.4    Systolic blood pressure 2023-08-01 2023-08-02 01:01:01
+                       reviewed         comment                    reviewer status
+      Adverse events.1       No     test review test_name (Medical Monitor)    new
+      Adverse events.2       No     test review test_name (Medical Monitor)    new
+      Vital signs.3         Yes another comment                  Reviewer 2    old
+      Vital signs.4          No                                                new
```
LDSamson, line 41: What makes these row names now appear?

LDSamson (commented):
Looks good indeed. Just a few minor questions

R/mod_review_forms.R
jthompson-arcus: Yes, a failsafe for testing although it might not actually be needed now. It keeps the module from crashing if `active_form()` isn't in `review_data`. Previously that would return an empty data frame.

tests/testthat/_snaps/mod_review_forms.md
jthompson-arcus: It comes from `rbind.data.frame(..., make.row.names = TRUE)`. If that argument is turned to `FALSE` it won't impute them. I thought about setting it to `FALSE` but I thought it was kind of nice that the output is capturing the list data structure.

R/mod_review_forms.R
LDSamson: Better failsafe than sorry ;)

tests/testthat/_snaps/mod_review_forms.md
LDSamson: I see. So the row names are not there in the original data r$review_data object? 
And are not transferred to the database, I assume. 
Thinking of keeping the r$review_data and the review database mostly the same

LDSamson: All good! that answered my question already

LDSamson (approved):
Looks good. I am fine with merging this one in #168 as you suggested

jthompson-arcus:
Will be merged into #168.

PR #171, opened 2025-02-27
Add merge_item_pairs functionality

### LINKED ISSUE #165: Add functionality to merge 'other' category of an item
In our case, we often have items that give additional explanation about an item, for example about why an item is missing, or why a drug is not administered. These columns are tightly coupled to one specific item, and are often empty unless certain conditions are met. 

It would be quite helpful if we add this functionality for us since it occurs often.
Basic idea is that we define which items should be paired to others in the metadata, something like this:

Image

And then a function will merge both items into one so that the value is something like:

Image

LDSamson:
Closes #165 .

The idea is that you can add a column named 'merge_with' in the metadata to define item variable pairs. The values in the variable with the name in `merge_with` will be added in brackets behind the value of the main variable.

image

When merging raw data with metadata, the result will be something like this:
```
 item_value                   
Other (1 time every 2 months)
Other (once)                 
Other (500-0-1000mg)         
Other (every 6 month)           
```

I am trying to keep this PR small because there are some big PRS open currently. But once these are merged, we can also add this `merge_with` column to the metadata, and update `clinsightful_data` with it.


jthompson-arcus (changes requested):
@LDSamson So I saw that there were several of these other fields in our raw data, so I went through the effort of updating `metadata` and `clinsightful_data` (which I think this PR should do anyways). This made the application crash at `R/fct_tables.R#332` which creates the medication table. That table expects the existence of these "other" columns for unit, frequency, and route.

I think logic in the app that expects these "other" columns to exist should be relaxed and I think `metadata` and clinsightful_data` should be updated to incorporate this merging capability.

LDSamson:
@jthompson-arcus thanks for checking. I think the issue that you describe is not an issue with the added code but just with the metadata because you are changing it. There is a reason why I did not do that yet, see last sentence in my explanation:

> I am trying to keep this PR small because there are some big PRS open currently. But once these are merged, we can also add this `merge_with` column to the metadata, and update `clinsightful_data` with it.

As soon as we implement it in out metadata, we can simply remove the hardcoded piece in the medication table. I guess I can work on that now. There will be merge conflicts though with other open PRs, we should be quite careful with that

Is there anything else that you think should be adjusted? Does the rest of the code look okay?

LDSamson:
@jthompson-arcus the metadata and study data is now updated

R/fct_merge_item_pairs.R
```diff
@@ -0,0 +1,189 @@
...
+    ...
+){
+  suffix <- suffix %||% ""
+  stopifnot(is.data.frame(data), is.character(suffix))
+  if (is.na(suffix) || suffix == ""){
+    cat("No suffix defined. Skipping merging of items.\n")
+    return(data)
+  }
+  cat("Using suffix '", suffix, "' to detect item pairs\n", sep = "")
+  other_vars <- unique(data$item_name[grepl(suffix, data$item_name)])
```
jthompson-arcus, line 126:
```suggestion
  other_vars <- unique(grep(suffix, data$item_name, value = TRUE))
```
Just a little more succinct.

jthompson-arcus: Sorry, meant to add this in a review comment.

jthompson-arcus (commented):
Looks pretty good to me. Just need to fix the merge conflicts.

LDSamson:
I merged the metadata files to a new one and recreated the internal datasets.
I also now applied the color scheme that is shown in our vignettes:
image

LDSamson:
@jthompson-arcus now also with updated .png snaps. I think it should be good to go now

jthompson-arcus (approved):
I like this PR. No other notes.

PR #172, opened 2025-02-27
Split review data

jthompson-arcus:


LDSamson (commented):
On closer inspection:

Previously, the contents of r$review_data was a `list`. Now we are expecting a `reactiveValues` object. I think our tests should reflect this design change, meaning that a `reactiveValues` object should also be used in the module tests.

I will implement this and then merge the PR

PR #175, opened 2025-03-04
Add helper utility function for creating custom metadata and config files

LDSamson:
@aclark02-arcus I felt like these helper functions can be useful. Let me know what you think of these. 

The files will be created in the current working directory by default and will be a copy of the metadata/config file in the package.

Note that the metadata.xlsx file should move to the `inst` folder for this function to also work without using `devtools::install()` which will probably cause merge conflicts.


R/fct_customize_clinsight_helpers.R
```diff
@@ -0,0 +1,61 @@
...
+  }
+  file.copy(template_path, config_path, overwrite = FALSE)
+  message(
+    "Creating a customizable ClinSight config file in the following location:\n", 
+    config_path, ".\n",
+    "To use this config file, save it in a custom location and use \n",
+    "    'Sys.setenv('CONFIG_PATH' = '", config_path, "')'\n", 
+    "to specify the config file location so that ClinSight can find it."
+  )
+  file.show(config_path)
```
aclark02-arcus, line 60: Hi @LDSamson, I would recommend using `file.edit()` here instead of `file.show()`. Only because `file.edit()` will open the file in the users R session. `file.show()` actually opened the file in VS (as seen below):

```suggestion
  file.edit(config_path)
```

image

aclark02-arcus (changes requested):
Hi @LDSamson, I like this PR. It will help some users get up & started more quickly...  which I think this package desperately needs! We'll have to be really thoughtful when we start composing the user guide.

I requested a small change below. Let me know your thoughts. While you are considering pushing any edits, you may want to update the NEWS file with a new bullet point introducing these new functions.

aclark02-arcus:
Looks good!

PR #179, opened 2025-04-02
Update chromote

LDSamson:
Resolves errors using latest Chrome version for shinytest2 tests (see https://github.com/rstudio/chromote/issues/204).

Note that using `renv::snapshot` also updated RSPM links of packages that were not updated; I don't mind that since it might prevent `renv` to build several packages from source, possibly speeding up the `renv::restore` process.


jthompson-arcus:
@LDSamson what do you think about specifying two snapshots as repos in the `renv.lock`. I think it might simplify some of the issues you were having on the GHAs.

LDSamson:
Closing this since the solution in #180 is cleaner.

PR #182, opened 2025-04-04
Add row review feature test

jthompson-arcus:


LDSamson (commented):
Hi, this is indeed quite in the right direction, thanks!

Few very quick comments:

In order to seamlessly integrate this in our traceability matrix, we need the tests in their own feature description block, e.g:

```
describe("Feature 3 | Save review per row. As a user, I want to be able to save a review of only specific rows in a table", {

  it("Scenario 1 - Save row review . etc.etc", {

  })
 it("Scenario 2 - e.g. Undo review of several rows. etc.etc.", {

 })
etc. etc.
})
```

Indeed, this will be slower since we need to close the test app and start a new one, but I think it is warranted here.

jthompson-arcus:
> Hi, this is indeed quite in the right direction, thanks!
> 
> Few very quick comments:
> 
> In order to seamlessly integrate this in our traceability matrix, we need the tests in their own feature description block, e.g:
> 
> ```
> describe("Feature 3 | Save review per row. As a user, I want to be able to save a review of only specific rows in a table", {
> 
>   it("Scenario 1 - Save row review . etc.etc", {
> 
>   })
>  it("Scenario 2 - e.g. Undo review of several rows. etc.etc.", {
> 
>  })
> etc. etc.
> })
> ```
> 
> Indeed, this will be slower since we need to close the test app and start a new one, but I think it is warranted here.

I feel like I both don't understand this and do. Why would this not be considered a "scenario" for "feature 2" which is saving a review?

LDSamson:
> > Hi, this is indeed quite in the right direction, thanks!
> > Few very quick comments:
> > In order to seamlessly integrate this in our traceability matrix, we need the tests in their own feature description block, e.g:
> > ```
> > describe("Feature 3 | Save review per row. As a user, I want to be able to save a review of only specific rows in a table", {
> > 
> >   it("Scenario 1 - Save row review . etc.etc", {
> > 
> >   })
> >  it("Scenario 2 - e.g. Undo review of several rows. etc.etc.", {
> > 
> >  })
> > etc. etc.
> > })
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Indeed, this will be slower since we need to close the test app and start a new one, but I think it is warranted here.
> 
> I feel like I both don't understand this and do. Why would this not be considered a "scenario" for "feature 2" which is saving a review?

Note that this is also how we discussed it in our meetings. There is no 100% right or wrong answer here, but I would feel like splitting it up is the best. One of the reasons is that the changes to the original code are substantial, this is a large additional feature. The previous feature 02 was aimed at testing reviewing an entire form at once (we could change the feature's name for clarity). Other reason is that by separating it, it is easier to focus on the new additions, both in the testing scripts and in the features/scenarios to be tested.

jthompson-arcus:
I did put it in its own feature. Is there anything else that needs to be included in this test initially?

tests/testthat/test-app_feature_05.R
jthompson-arcus, line 1: @LDSamson do you think there's a value add to making this a partial undo? Like maybe undoing the review of the second record.

LDSamson: Makes sense since that partial undo is unique to the row_review feature

jthompson-arcus:
@LDSamson merged your additional tests in. Think we are good to go here.

PR #183, opened 2025-04-07
Branch dedicated to `dev` demo

### LINKED ISSUE #176: Create shinyapps.io demo app deployment

aclark02-arcus:
Currently, not operational when using push button deployment from Rstudio IDE, with the following error:

image

Probably because a different version exists on pkg manager??? As seen here:

image


LDSamson:
1.84.0-0 is the BH version of RSPM date 2024-09-17. 
image

The link with date 2025-04-01 is named RSPM2 in the `renv.lock` file, not RSPM. I don't know why it does not recognize the correct source link here.

manifest.json
```diff
@@ -31,104 +31,104 @@
...
         "Encoding": "UTF-8",
-        "Built": "R 4.3.0; ; 2024-01-23 01:39:34 UTC; windows"
+        "Built": "R 4.4.0; ; 2024-04-25 18:45:03 UTC; windows"
       }
     },
     "BH": {
-      "Source": "RSPM",
-      "Repository": "https://packagemanager.posit.co/cran/2023-11-01",
+      "Source": "RSPM2",
+      "Repository": "https://packagemanager.posit.co/cran/2025-04-01",
```
LDSamson, line 39: Here you are using a different source link for the BH package. It should be 2024-09-17. Could that be the source of your error?

aclark02-arcus: Just so we're all on the same page, it's my understanding that `manifest.json` is only used by Posit Connect. When deploying to shinyapps.io, Posit doesn't acknowledge this user created file, generated by `rsconnect::writeManiest()`. 

Yet Posit does automatically build it's own manifest for deployment. I confirmed this by deleting the `manifest.json` file and attempting to re-deploy and I still got the same error.

So, in summary, `rsconnect::writeManiest()` is likely capturing the same symptom that shinyapps.io is running into.

jthompson-arcus: Can you share more of your workflow? `rsconnect::writeManifest()` doesn't work out of the box with `renv` profiles, so I'm just curious if there's something else going on as well.

jthompson-arcus: I wouldn't think `rsconnect` and dependencies would need to be added to `renv.lock` either. Did you do that just for transparency or was it needed?

aclark02-arcus: > Can you share more of your workflow? `rsconnect::writeManifest()` doesn't work out of the box with `renv` profiles, so I'm just curious if there's something else going on as well.

see `app.R` on this branch.

aclark02-arcus:
Deployments dedicated to `v0.3.0`:
- https://rinpharma.shinyapps.io/clinsight_v030/
- https://rinpharma.shinyapps.io/clinsight

aclark02-arcus:
Close with [the deployment](https://github.com/openpharma/clinsight/pull/275) of `v0.4.0`

PR #185, opened 2025-04-09
Add test clinsight function

LDSamson:
Continuing the discussion started [here](https://github.com/openpharma/clinsight/pull/181#discussion_r2029048796) in a separate PR.

Main goal of the `test_clinsight()` function is to easily run ClinSight with custom data and metadata, make some changes and run it again. It is not meant for use in production. 

I did not yet remove the `clinsight_config` argument yet because of two reasons:

1. I feel like it can also be useful to test configurations that are not related to the location of the data, for example the `allow_to_review` or `allow_listing_download` settings.
2. We already have a data_folder argument in the `run_app()`. This argument's purpose is to customize the path in which all clinsight data is stored and, if used, ignore any file paths specified in the `golem-config.yml` file.

The purpose of the `data_folder` argument within `run_app` is made clearer in the documentation, and a message will be shown when it is used, stating that any data path in the`golem-config.yml` file file will be ignored. This was always the intended behavior but previously would not have worked when a custom file path was specified in the `golem-config.yml` file.

 @Jeff-Thompson12 I feel like this solves any issues with being able to set the `clinsight_config` argument within `test_clinsight`, since it will just work with any settings, but I'd love to hear your thoughts.

Lastly, I think this function can probably replace all the `load_and_run_app()` functions that we are using throughout the application for testing purposes.


R/fct_test_clinsight.R
```diff
@@ -0,0 +1,52 @@
...
+    clinsight_config == "default" & 
+    (!identical(clinsight_data, clinsightful_data) | !identical(meta_data, metadata))
+  ){
+    stop("The 'default' config should not be used with custom data.")
+  }
+  
+  temp_folder <- tempfile(tmpdir = tempdir())
+  dir.create(temp_folder, recursive = TRUE)
+  saveRDS(clinsight_data, file.path(temp_folder, "study_data.rds"))
+  saveRDS(meta_data, file.path(temp_folder, "metadata.rds"))
```
jthompson-arcus, line 41: We should grab the `basename()` from the configuration file being used. This assumes that a user will always call these `study_data.rds` and `metadata.rds`.

R/run_app.R
```diff
@@ -34,10 +35,16 @@ run_app <- function(
...
-    if(is.character(data)) data <- file.path(data_folder, data)
-    if(is.character(meta)) meta <- file.path(data_folder, meta)
-    user_db <-  file.path(data_folder, user_db)
-    credentials_db <- file.path(data_folder, credentials_db)
+    golem::cat_dev(
+      "Custom folder path provided in the 'data_folder' argument.",
+      "File paths specified in the config.yml will be ignored."
+    )
+    if(is.character(data)) data <- file.path(data_folder, basename(data))
+    if(is.character(meta)) meta <- file.path(data_folder, basename(meta))
```
jthompson-arcus, line 43: There should be some console indication that only the `basename()` is being grabbed from the configuration.

jthompson-arcus (changes requested):
@LDSamson I think I see your vision a little bit more. I do have a few comments below.

R/run_app.R
LDSamson: @jthompson-arcus Do you mean that the lines 38-41 are not sufficient?

jthompson-arcus: My bad. It's obviously there.

R/fct_test_clinsight.R
```diff
@@ -28,17 +28,21 @@ test_clinsight <- function(
   stopifnot("Expecting a data frame" = is.data.frame(clinsight_data))
   stopifnot("Expecting a list" = inherits(meta_data, "list"))
   stopifnot(is.character(clinsight_config))
+  study_data_path <- get_golem_config("study_data", config = clinsight_config)
+  meta_data_path <- get_golem_config("meta_data", config = clinsight_config)
   if (
-    clinsight_config == "default" & 
-    (!identical(clinsight_data, clinsightful_data) | !identical(meta_data, metadata))
+    clinsight_config %in% c("default", "dev") | 
+    !is.character(study_data_path) | !is.character(meta_data_path)
```
jthompson-arcus, line 35: @LDSamson doesn't look like you are checking for custom data anymore here

jthompson-arcus: Okay, I was a little confused with what you were checking earlier. Makes a little more sense now.

jthompson-arcus (approved):
LGTM

PR #186, opened 2025-04-10
add more row review tests

LDSamson:


LDSamson:
@jthompson-arcus I added two more scenarios so that more code will be tested. Just merge in your branch if you are okay with. 
If you see any place for improvements, please go ahead and add these.

jthompson-arcus (approved):
This all seems good to me.

PR #188, opened 2025-04-14
Try using multiversion website

LDSamson:
See https://github.com/insightsengineering/r-pkgdown-multiversion/tree/main

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -48,7 +48,7 @@ jobs:
       
       - name: R pkgdown Multi Version docs
         uses: insightsengineering/r-pkgdown-multiversion@v3.0.3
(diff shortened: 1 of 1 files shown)


LDSamson:
On further inspection, there might be some hidden issues with this approach. For example, for admiral they switched to an easier approach: https://github.com/pharmaverse/admiral/issues/2103
However, with their latest approach, there are also some issues with showing old versions on the site: https://github.com/pharmaverse/admiralci/issues/189

Other option would be an approach like this: https://github.com/Appsilon/rhino/pull/575

Multi version documentation might be implemented directly in `pkgdown` in the future: https://github.com/r-lib/pkgdown/issues/1373

LDSamson:
Closing this for now. No Time to implement this currently

PR #190, opened 2025-04-14
Fix table view row order when toggling `show_all`

### LINKED ISSUE #189: Toggling `show all participants` does not sort the interactive table intuitively
As discussed here: 

https://github.com/openpharma/clinsight/pull/146#discussion_r1915878919

Can be solved by either sorting on the server side or on the client side. Not sure what is favorable in this case.

LDSamson:
Fixes #189, an issue that @aclark02-arcus brought up in the past (see https://github.com/openpharma/clinsight/pull/146#discussion_r1915878919).

toggle_show_all_subjects-ezgif com-video-to-gif-converter

@aclark02-arcus @jthompson-arcus thoughts on this? Would this be a big overhead? Would this be a problem in really large tables? If so, is there a better alternative?

I can update the documentation etc in this branch once we all think this is indeed the best way forward.


LDSamson:
When performing a `microbenchmark` test, the overhead does look negligible for small datasets (tested on appdata[["Medication"]] ):

```
Unit: milliseconds
            expr          min      lq     mean   median       uq      max neval
    with subject sorting 76.8537 84.2463 102.1257 94.71300 107.2309 764.9813  1000
 without subject sorting 76.1310 84.7082  99.5094 94.25035 106.3751 551.5502  1000
```

But I don't know what it would be on a large dataset.

aclark02-arcus (changes requested):
Hi @LDSamson,

This looks like a step in the right direction. Although, I'm curious why the sort isn't applied to all subjects displayed? The "active subject" is retained at the top which is a huge win, but all the other subjects are still be jumbled about in the table. See AE example below.  Wouldn't it be nice to see all patient's sorted by default? I am not a medical monitor, but I'd think it'd be convenient to scroll through each patient's block of rows when reviewing other AEs in the table. Would be easier to digest (for me).

image

PR #191, opened 2025-04-14
Create deploy renv profile

jthompson-arcus:


LDSamson (changes requested):
My only issue is that the naming is a bit confusing because we now have two profiles for deployment, namely `deploy` and `minimal`, which are both used for a different kind of deployment but that is not obvious from the name. 

My suggestion would be to rename it to something like `deploy_posit` and `deploy_minimal`.

NEWS.md
```diff
@@ -34,6 +34,7 @@
 - Added a more recent repo snapshot for `chromote` v0.5.0 used in Shiny tests
 - Added two helper functions (`create_clinsight_metadata()` and `create_clinsight_config`) to create custom `ClinSight` metadata and config files.
 - Added a feature test for row level review.
+- Added a "deploy" profile for deployments. Includes additional dependencies not included in the "minimal" profile for deployments utilizing `shinymanager` or the `app.R` file, while being more restrictive than the "full" profile.
```
LDSamson, line 37: Note that `shinymanager` is not included in the new profile. If you want that, you should add it to the dependencies in the `Description` file.

jthompson-arcus:
> My only issue is that the naming is a bit confusing because we now have two profiles for deployment, namely `deploy` and `minimal`, which are both used for a different kind of deployment but that is not obvious from the name.
> 
> My suggestion would be to rename it to something like `deploy_posit` and `deploy_minimal`.

I could see naming it `deploy_full` but that's awkward. Honestly this should be called the `full` profile and our current `full` profile should be called `dev`.

NEWS.md
jthompson-arcus: Yeah, overlook on my part

LDSamson:
> > My only issue is that the naming is a bit confusing because we now have two profiles for deployment, namely `deploy` and `minimal`, which are both used for a different kind of deployment but that is not obvious from the name.
> > My suggestion would be to rename it to something like `deploy_posit` and `deploy_minimal`.
> 
> I could see naming it `deploy_full` but that's awkward. Honestly this should be called the `full` profile and our current `full` profile should be called `dev`.
> 
> Also, I personally think `deploy` is more descriptive than `deploy_posit` since this profile is needed for basically every other deployment strategy.

All deployments in which the clinsight package is simply installed do not need the additional development packages and dependencies such as `pkgload`. For example, installing from a locally deployed package manager in which validated version of `clinsight` are installed. In general, I feel like it makes more sense to deliver production-ready applications in highly regulated environments not as source code but rather as bundled packages.

But regardless of that subjective point, my main point is that it should be clear that `minimal` is also used for deployment. If you create a profile called `deploy`, this is unclear

LDSamson (changes requested):
I am a bit confused, because now there is a `minimal` but also a `deploy_minimal` profile. The idea was only renaming it, right?

jthompson-arcus:
> I am a bit confused, because now there is a `minimal` but also a `deploy_minimal` profile. The idea was only renaming it, right?
> 
> grafik

@LDSamson the same thing happened the last time we changed profile names. You probably didn't notice because you were the one changing the name. I only have the three folders now. You can delete the "minimal" one to make it go away. I think it's because Git tracks files not really folders.

LDSamson (approved):
LGTM

Issue #144, opened 2024-11-21
Re-assess exported functions


jthompson-arcus:
@LDSamson @aclark02-arcus @jthompson-arcus

Should function be exported (Y/N)?

Function | Leon (Y/N) | Aaron (Y/N) | Jeff (Y/N) |
--------- | :-----------: | :------------: | :---------: |
`add_missing_columns()` | N | | N | 
`adjust_colnames()` | N | | |
`app_ui()` | N | | N |
`bind_rows_custom()` | N | | N | 
`check_appdata()` | N | | N |
`clean_dates()` | N | | N |
`collapse_column_vals()` | N | | Y Needed for a vignette |
`collapse_fct_levels()` | N | 
`collect_query_data()` | N |||
`cols_to_char()` | N | | | 
`create_report()` | N | | |
`create_table()` | Y Probably good to export so that users can create their own methods if needed | | ? This function only makes sense for outputs from `get_appdata()`. Having said that, do we have to export to allow users to create their own S3 methods?|
`create_unique_id()` | N | 
`custom_plot_theme()` | N | | |
`datatable_custom()` | N | | N |
`date_cols_to_char()` | N | | N |
`db_create()` | Y | | | 
`db_get_query()` | Maybe? |
`db_get_review()` | Maybe? | 
`db_save()` |  N | | 
`db_save_review()` | N | 
`db_slice_rows()` | N | | | 
`db_temp_connect()` | N | | |
`db_update()` | Y | | |
`expand_columns()` | N | | N |
`fig_barplot()` | N | | | 
`fig_boxplots()` | N | | | 
`fig_timeline()` | N | | | 
`fig_timeseries()` | N | | |
`fix_multiple_choice_vars()` | N | | | 
`get_appdata()` | N Not yet needed. Can always be changed back in the future. *EDIT: Export is useful. Useful for data pre-processing steps icm `create_table`, and probably good to make the app work with this data format, to speed up start up time* | | | 
`get_available_data()` | N | | | 
`get_base_value()` | Y (Can be useful for pre-processing data) | | | 
`get_db_connection()` | Y |||
`get_ggplot_layer_names()` | N | | N | 
`get_max_time()` | N | 
`get_meta_vars()` | N | | N | 
`get_metadata()` | Y | | |
`get_raw_csv_data()` | Y | | Y |
`get_review_data()` | N | | |
`get_static_overview_data()` | N | | |
[...]

PR #192, opened 2025-04-15
Un-export Functions

jthompson-arcus:
First pass at #144


jthompson-arcus:
@LDSamson one of the vignettes needed `collapse_column_vals()` so it is not dropped from the exported list.

PR #193, opened 2025-04-16
Fill out connect deployment documentation

jthompson-arcus:


jthompson-arcus:
@LDSamson I added the README render job to this branch. Works like a charm.

LDSamson (approved):
@jthompson-arcus not a big issue (I think) but still worth to mention: it does not use our development environment, thus it is not using the R version and Rmarkdown version that we are using in development (R 4.4.1).

jthompson-arcus:
> @jthompson-arcus not a big issue (I think) but still worth to mention: it does not use our development environment, thus it is not using the R version and Rmarkdown version that we are using in development (R 4.4.1).

We can adjust it. I just copy-pasted from another project. Didn't put much thought in it beyond that.

PR #198, opened 2025-04-23
Add form level review feature

### LINKED ISSUE #22: Allow for alternative review workflow - by form

LDSamson:
We now also have the need for this kind of functionality for a different study.

See below for the first attempt. study forms:
form_level_review_study_forms-ezgif com-video-to-gif-converter

Common forms:
form_level_review_common_forms-ezgif com-video-to-gif-converter

Still need to fine tune it, update tests and ensure that it is robust, but the basic functionality is there. @jthompson-arcus @aclark02-arcus thoughts?


jthompson-arcus:
@LDSamson First impression is that I like this. It did take me a while to understand what this PR was actually implementing though. I think it might be helpful to add tooltips to the toggle and I would probably move the toggle above the overall review checkbox.

Also, do we want to consider adding a modal confirmation on the review? Not exactly sure how we would implement it, but it's kind of crazy that a user can mark all records in a form as "reviewed" with 3 clicks.

LDSamson:
@jthompson-arcus I agree with both of the points you mentioned, we need it to be more user friendly, but I wanted to have some first opinions already. Good to hear you think it is useful.

I plan to add a tooltip, maybe some extra text in the review panel on the left next to the review mode toggle, a notification of the change in the top left corner showing the change in review mode, and indeed some sort of confirmation when a user is about to review all records of a form at once.

Ideally we would have an undo button to go back to the exact previous state, but that might be a bit difficult (maybe for a follow-up PR, if ever)

LDSamson:
See below for the improvements, as discussed:
review_typeimprovements

I plan to improve the message in the modal but I ran out of inspiration today. 
@jthompson-arcus @aclark02-arcus thoughts so far on the UI and the additional tooltips/messages? 

I tried the new button on the top but I feel like it is then a bit in the way of the standard review actions that will be used much more than this button.

To do: 
- tweak the modal message
- add unit/integration tests

LDSamson:
"Show all participants" is disabled in `form` review mode and figures are hidden.

LDSamson:
@jthompson-arcus should be ready now.

It's quite complicated to get the reactivity right! But I think I've got it. Because of this, I added many tests and test scenarios.

I tried to make the added snapshots quite focused, so that they are less likely to change in the future.
We could do with less snapshots in hindsight. Let me know what you think of it.

@aclark02-arcus feel free to also try out this new functionality and/or comment on it!

R/app_server.R
```diff
@@ -200,6 +201,11 @@ app_server <- function(
     bslib::nav_select(id = id_to_change, selected = navinfo$active_form)
   })
   
+  output$form_level_review <- reactive({
+    identical(session$userData$review_type(), "form")
+  })
+  outputOptions(output, "form_level_review", suspendWhenHidden = FALSE)
+
```
LDSamson, line 208: Needed for 1) being able to smoothly change the UI based on the review type, and 2) for testing purposes

R/mod_review_form_tbl.R
```diff
@@ -69,26 +69,26 @@ mod_review_form_tbl_server <- function(
...
-    
     observe({
       golem::cat_dev(form, "| Resetting userData\n")
       reload_data(reload_data() + 1)
       datatable_rendered(NULL)
       session$userData$pending_form_review_status[[form]] <- NULL
       session$userData$pending_review_records[[form]] <- data.frame(id = integer(), reviewed = character())
-    }) |> 
-      bindEvent(active_subject(), form_review_data(), form_data())
+    }, priority = 1) |>
```
LDSamson, line 87: Priority needed to be set, otherwise sometimes this would run AFTER the data was rendered again, which would set datatable_rendered to NULL unintentionally.

jthompson-arcus: Do you want to do something like set it to 100? Should always be the first thing to trigger.

LDSamson: Sure, we can do that
```suggestion
    }, priority = 100) |> 
```

LDSamson:
@jthompson-arcus I noticed that, when in form review mode and only reviewing a few rows, the warning message should either be removed or tailored to the user. I can work on that tomorrow or wednesday

LDSamson:
@jthompson-arcus the issue that I mentioned last week is now resolved. Additional tests were added.

R/mod_review_forms.R
```diff
@@ -113,21 +124,38 @@ mod_review_forms_server <- function(
...
     
     review_data_active <- reactive({
-      tryCatch(
-        subset(r$review_data[[active_form()]], subject_id == r$subject_id),
+      tryCatch({
+        df <- subset(r$review_data[[active_form()]]) 
+        if (!identical(input$review_type, "form")){
+          df <- subset(df, subject_id == r$subject_id)
+        }
+        df
```
jthompson-arcus: I think it might be easier to read if we wrote it like this `subset(r$review_data[[active_form()]], identical(input$review_type, "form") | subject_id == r$subject_id)`

R/mod_review_forms.R
```diff
@@ -113,21 +124,38 @@ mod_review_forms_server <- function(
...
+    
     review_indeterminate <- reactiveVal()
     
     observeEvent(review_indeterminate(), {
       shinyjs::runjs(sprintf("$('#%s').prop('indeterminate', %s)", ns("form_reviewed"), tolower(review_indeterminate())))
     })
     
     observe({
-      req(session$userData$pending_review_records[[active_form()]])
+      req(session$userData$pending_review_records[[active_form()]], review_data_active())
```
jthompson-arcus: What situation required this addition?

R/mod_study_forms.R
jthompson-arcus, line 1: I'm not psyched that you can still view the subject level plots when viewing the full form. But maybe that can be a follow up PR. The row review never fully integrated with the plots anyways.

jthompson-arcus (commented):
@LDSamson I don't really have to bandwidth to go through the tests. I did have a few comments below but nothing major. Seems to be working as expected and definitely a welcome feature addition.

R/mod_review_forms.R
LDSamson: I agree. The first subset doesn't even do anything now

R/mod_review_forms.R
LDSamson: We can probably do without. I don't think it is essential here. I think it is a leftover of me trying to fix the reactivity. I will check it

R/mod_study_forms.R
LDSamson: There is currently an issue with the id for the switch between plot and table view which I think should be addressed in a separate issue/PR. The switch should also be disabled if all data is continuous but it is not.

image

R/mod_review_forms.R
LDSamson: I removed it.

R/mod_review_forms.R
LDSamson: It's now updated as you suggested

LDSamson:
@jthompson-arcus thanks for the review. If you want, we can go over the tests in a meeting to discuss on a global level what they do and what scenarios they address.

jthompson-arcus (changes requested):
Looks like version number isn't incremented. Maybe the last push to `dev` didn't get merged in.

jthompson-arcus (approved):
LGTM

PR #207, opened 2025-05-12
Remove 'Name' column requirement for common forms

### LINKED ISSUE #52: "MH Name" `item_name` is required in 'common_forms' tab in metadata.xlsx
Hi,

Just wanted to point out that thanks to the following chunk of code, `MH Name` is required, so this should be documented in the package. Notice that the "MH" is stripped so the variable is just called "Name" here:

https://github.com/openpharma/clinsight/blob/f29a720bd6a705241adf76843a4fc5195a51c7c0/R/fct_tables.R#L308

LDSamson:
Closes #52.

Until now, `common_forms` always need a column named 'Name' in their table created with the `clinsight::create_table()` function.

This is because common forms are expected to have a common item (`Name`) to which all the other items of that row belong to (think of the name of adverse events). 

This PR removes this requirement and also allows tables in the common_forms tab if they do not have a 'Name' column. Additionally, if forms in study_forms have such a name column, this name will be seen as the identifier of the table row (thgus, more like previous common_form behaviour). This affect on which item something will be queried in the query system.

Thus, this makes the system more flexible, and you can change forms like this if you want to (EDIT: the example does not really make sense, but it is just for illustration purposes):
image


aclark02-arcus (approved):
This looks good to me, but in the future, can you please submit more minimal PRs? This one took a long time to review.

Nah, but in all seriousness, this is a welcome change. As mentioned previously, we encountered this issue early on as a lot of our forms didn't have a natural "Name" column. But regardless, we chose chose one that was "good enough" and manually tracked them in our metadata using a variable called `form_level_id_var`. And instead of just renaming the variable in the raw data, we were doing it here, in `get_available_data()` with this little code chunk:


image

Notice the `map()` by form_names... you don't have that feature yet so you can ignore. It's related to displaying more than one form per tab (item_group). So I'll have to weigh whether to keep this code or not. Technically it still has some usefulness!

LDSamson:
Good that it works! Noted, I will try to minimize my PRs in the future, this is too much  ;)

We should explore if we can just remove the suffixes before the `Name` variable ('AE', 'CP' etc). That would make it easier to maintain.

PR #214, opened 2025-05-20
Improve branding

### LINKED ISSUE #213: Improve branding
Opening an issue to work on three things:

1. Improving `ClinSight` logo so that it appears more prominently, like a title. Right now, it is very dainty and small.
2. Add a favicon & study info to the browser tab
3. If study logo is available, it could be added in place of the `meta$study_name` location boxed below

Image

aclark02-arcus:
Seeks to close #213 

Specifically:
1. Improving `ClinSight` logo so that it appears more prominently, like a title.
    - Right now, it is very dainty and small. Created a mock-up placeholder logo until a more official one is generated by GCP.
    - updated the "ClinSight" logo's hyperlink to point to the `pkgdown` docs site.
2. Add a favicon & study info to the browser tab
    - Used the magnifying glass w/ GCP logo in the center
    - When study_name exists, it will show up in the tab - very useful imho
3. If study logo is available, it could be displayed in leu of a textOutput()
    - Our org has beautiful logos for each study, so we want to use those if possible


aclark02-arcus:
FYI - sharing code that Leon passed along to me to test adding a study_name more easily:
```r
if(rlang::is_interactive()){
  raw_data_folder <- app_sys("tests/testthat/fixtures/testapp-raw/data1pt")
  # meta_path <- testthat::test_path("fixtures/testapp-raw/metadata_fake_study.xlsx") 
  # meta_path <- testthat::test_path("fixtures/testapp-raw/metadata_study_example.xlsx") 
}

load_and_run_app <- function(){
  temp_folder <- tempfile(tmpdir = tempdir())
  dir.create(temp_folder)
  old_golem_config <- Sys.getenv("GOLEM_CONFIG_ACTIVE")
  Sys.setenv("GOLEM_CONFIG_ACTIVE" = "test")
  meta <- get_metadata(meta_path)
  merged_data <- raw_data_folder |> 
    get_raw_csv_data(synch_time = "2024-01-01 00:00:00") |> 
    merge_meta_with_data(meta = meta)
  saveRDS(merged_data, file.path(temp_folder, "study_data.rds"))
  saveRDS(meta, file.path(temp_folder, "metadata.rds"))
  run_app(
    data_folder = temp_folder,
    onStart = \(){onStop(\(){
      unlink(temp_folder, recursive = TRUE); 
      Sys.setenv("GOLEM_CONFIG_ACTIVE" = old_golem_config)
    })}
  )
}

**Comment by aclark02-arcus** [comment]
Okay, so there are three scenarios to test for the new study name display that correspond different metadata.xlsx files.

First, using `golem::run_dev()`, we test if the `meta$study_name` is `NULL`, then I have the app displaying an org logo by default, if it exists. I added GCPs to show. So if an org wants to replace that, they just need to replace it with their own org logo in the `study_logos/` dir. Also notice the new pretty favicon and the browser tab just says "ClinSight Study".

image

Next, if the deployer sets up `meta$study_name` value with the name of a study that doesn't have a logo associated with it, the app displays the text found in the workbook.

image

Last, if the deployer sets up `meta$study_name` value with the name of a study that **does indeed** have a logo associated with it, the app displays that logo as long as the name of the png matches the name of the study. The png syntax does have to be all lower case.

image

@LDSamson, thoughts?

File: clinsight_metadata.xlsx
Comment (LDSamson) [inline clinsight_metadata.xlsx:1]: I don't think this file should have been committed, right? If so, it is not the right place for it, it should be in the inst folder. What did you change exactly here?

File: R/app_ui.R [hunk 1]
```diff
@@ -20,8 +21,9 @@ app_ui <- function(request){
         ),
       bg = "#43464c",
       title = tags$a(
-        href = "/",
-        tags$img(src='www/logo_in_app_w_margin.png', height = '45')
+        href = "https://openpharma.github.io/clinsight/",
+        target = "_blank",
+        tags$img(src='www/logo_in_app_w_margin_wide.png', height = '40')
```
LDSamson: I see what you are trying to do here, but `href = "/"` is quite useful on our production server.

Our setup is something like "https://our_domain.com/study_1, https://our_domain.com/study_2" etc. Clicking on the logo now brings us back to the home page ("https://ourt_domain.com"), showing the list of all studies/ all deployed applications.

Therefore, we would like to keep it that way for now or at least not hard code this. Maybe we can implement the link to the manual somewhere else?

dev/app.R
```diff
@@ -37,3 +37,30 @@ run_app(
...
+  saveRDS(merged_data, file.path(temp_folder, "study_data.rds"))
+  saveRDS(meta, file.path(temp_folder, "metadata.rds"))
+  run_app(
+    data_folder = temp_folder,
+    onStart = \(){onStop(\(){
+      unlink(temp_folder, recursive = TRUE); 
+      Sys.setenv("GOLEM_CONFIG_ACTIVE" = old_golem_config)
+    })}
+  )
+}
```
LDSamson: Bear in mind that this is exactly the code in tests/testthat/fixtures/testapp-raw/app.R.

Maybe it would be better to refer to that script instead, because that is automatically tested; it prevents us from needing to update this script in too many places in the future.

R/app_ui.R
```diff
@@ -42,7 +44,8 @@ app_ui <- function(request){
       ),
       bslib::nav_spacer(),
       bslib::nav_item(
-        tags$h3(textOutput("study_name"), class = "text-secondary")
+        shiny::uiOutput("study_name_display")
+        # tags$h3(textOutput("study_name"), class = "text-secondary")
```
LDSamson:
```suggestion
        uiOutput("study_name_display")
```

R/app_ui.R
```diff
@@ -73,7 +76,7 @@ golem_add_external_resources <- function() {
     favicon(),
     bundle_resources(
       path = app_sys("app/www"),
-      app_title = "clinsight"
+      app_title = paste("ClinSight", study_name)
```
LDSamson:
```suggestion
      app_title = paste0("ClinSight | ", study_name)
```

R/app_server.R
```diff
@@ -278,14 +278,35 @@ app_server <- function(
...
     
-    output[["study_name"]] <-  renderText({
+    output[["study_name_display"]] <-  renderUI({
+      # If study_name is NULL, Set it to an empty string
       study_name <- meta$settings$study_name %||% ""
-      if (nchar(study_name) > 40){
-        paste0(trimws(substr(study_name, 1, 37)), "...")
+      
+      # create potential path to study_logo file so we can search for them
+      study_logo <- paste0('www/study_logos/',tolower(meta$settings$study_name %||% "org_logo"),".png")
```
LDSamson: It would be great if this can also work if we install the package, and provide a link to a study_logo folder. Do you think that is possible? Maybe by setting the path in the metadata? If we follow our SOPs we should not change the source code of a validated package.

tests/testthat/fixtures/testapp-raw/metadata_fake_study.xlsx
LDSamson, line 1: Are these new metadata .xlsx files needed? Maybe its a bit overkill and also hard to maintain in the future? See the tests that are already implemented for testing the study name

LDSamson (changes requested):
Some initial thoughts, see comments

dev/app.R
aclark02-arcus: Sure, I didn't know that. I was just putting it here temporarily for testing purposes. I will delete.

R/app_ui.R
aclark02-arcus: @LDSamson, yes I thought about putting some sort of separator here, but that may not be helpful when `meta$study_name` is `NULL`, since "Study" will be used by default. So with the vertical var, it will display "ClinSight | Study" instead of "ClinSight Study". Is that okay?

tests/testthat/fixtures/testapp-raw/metadata_fake_study.xlsx
aclark02-arcus: Yes, I just created them for initial testing so the PR reviewer didn't have to make them. I was planning to delete at some point. I will leave them there for now.

R/app_server.R
aclark02-arcus: I suppose we could be more explicit with where the study logo exists, instead of just assuming it exists in the `www/` folder. So, you envision this being another field in the `meta$settings`?

clinsight_metadata.xlsx
aclark02-arcus: Yep, committed in error. removed.

R/app_server.R
LDSamson: My philosophy with ClinSight is that, ultimately, we can use the package with any study without changing the source code, therefore study-specific assets should be separated from the source, or we should at least be able to do so. 

Do you keep data of all studies in the same folder? We keep it all separate. If you also do that: why not just fix the logo name and do something like this:

- Add a folder named `assets` (or similar) to the resource paths
- Look for `study_logo.png` in this folder
- if it is there, then show it.

If you want to display the organisations' logo instead, just name it `study_logo.png` and place it there. 
We could make it more flexible by providing the location of the assets folder in the `golem-config.yml`. I think that would be the right place instead of the metadata since the path probably can stay the same for multiple studies (if you are able to use relative paths).

@aclark02-arcus thoughts?

aclark02-arcus: Ah, yes that makes more sense. I could work up something in another commit. I'll let you know if I have any follow-up questions.

R/app_ui.R
aclark02-arcus: @LDSamson Any more thoughts on this? I could change the default `study_name here to be "Unknown Study", but that seems a bit "extra". "ClinSight Study" seems to make sense, but I think "ClinSight | Study" may be confusing. An alternative is we could remove any suffix when `meta$settings$study_name` is `NULL` so that the tab just says "ClinSight".

R/app_server.R
aclark02-arcus: Hi @LDSamson, I pushed a couple of commits on this topic. Please let me know if my interpretation of your comments was heading in the right direction. I did leave the new golem config field "blank" which automatically renders as `NULL`. Not sure how you feel about that.

R/app_ui.R
LDSamson: I think my preference would be your last suggestion, to keep the separator and just remove any suffix if the study_name is NULL. But it is a bit subjective of course.

R/app_ui.R
```diff
@@ -64,7 +65,19 @@ app_ui <- function(request){
...
 #' @noRd
-golem_add_external_resources <- function() {
+golem_add_external_resources <- function(study_name) {
+  
+  # If a study asset path is provided, verify it exists before adding it as a 
+  # resource path
+  asset_path <- get_golem_config("study_asset_path")
+  if(!is.null(asset_path)){
+    if(dir.exists(asset_path)) {
+      add_resource_path("assets", asset_path)
```
aclark02-arcus: Adding 'assets' to the resource path doesn't seem to be working for me, FYI.

aclark02-arcus (commented):
One comment on this approach. I don't see why we couldn't just remove this code and use `get_golem_config('study_asset_path') directly in `app_server.R`'s `renderUI()`, do you @LDSamson

R/app_ui.R
LDSamson: what error do you get? For me it works fine

LDSamson:
> One comment on this approach. I don't see why we couldn't just remove this code and use `get_golem_config('study_asset_path') directly in `app_server.R`'s `renderUI()`, do you @LDSamson

As far as I understand you should look up the resource directly in any server side logic, not by using the "assets" keyword. The resource path named 'assets' is meant for use client side (when the link is declared in the UI). See also here: https://engineering-shiny.org/golem.html#instappwww. 

It could be that an assets folder is not needed if you run it completely on the server side. But I am not 100% sure if an image tag would work in that case. Disadvantage of running it on the server side is that the logo shows up fairly late though, only when all the loading is done. Is that acceptable do you think?

We can discuss, I don't know what is best. I can create an example of how to run it fully client side. You should then probably place the logo verification logic in `run_app`.

R/app_ui.R
aclark02-arcus: No error @LDSamson, but the code can't find the image file, so it just displays the study name as text.

aclark02-arcus:
> not by using the "assets" keyword

Good to know! Thanks for the resource.

> Disadvantage of running it on the server side is that the logo shows up fairly late though, only when all the loading is done. Is that acceptable do you think?

It's slightly annoying, but not a big deal IMHO.

> I can create an example of how to run it fully client side.

I would be interested in this. I've spent a little more time than I'd like to on this PR, so if you have some ideas, please feel free to share / contribute. You can push directly to this branch if you want.

LDSamson:
> > not by using the "assets" keyword
> 
> Good to know! Thanks for the resource.
> 
> > Disadvantage of running it on the server side is that the logo shows up fairly late though, only when all the loading is done. Is that acceptable do you think?
> 
> It's slightly annoying, but not a big deal IMHO.
> 
> > I can create an example of how to run it fully client side.
> 
> I would be interested in this. I've spent a little more time than I'd like to on this PR, so if you have some ideas, please feel free to share / contribute. You can push directly to this branch if you want.

See the other PR I created for the example. It's based on your PR: https://github.com/openpharma/clinsight/pull/215

R/app_server.R
aclark02-arcus: stale convo

R/app_ui.R
aclark02-arcus: Will re-deploy with recent merge of #215... stay tuned.

aclark02-arcus: This looks good on our end.

tests/testthat/fixtures/testapp-raw/metadata_fake_study.xlsx
aclark02-arcus: removed.

aclark02-arcus:
@LDSamson, after merge of #215 and a little R CMD Check cleanup, this one is ready to go. If it looks good, please merge.

inst/app/www/logo_mag_for_favicon.png
LDSamson, line 1: Probably not needed anymore, right?

LDSamson (changes requested):
Looks good. One comment: you performed extensive manual testing for this. Can't we capture this manual testing in an official test, as discussed previously? I think it should be fairly straightforward by adding a new scenario for the app-feature_04 test: 

- Add a non-standard asset folder in the [`testapp-raw` folder](https://github.com/openpharma/clinsight/tree/dev/tests/testthat/fixtures/testapp-raw) (for example, `assets_custom`) with the test image that you used for testing
- Add a custom `config.yml` file in the same folder. Use the `test` config.yml configuration, but adjust the `study_logo_path` to the new one (`assets_custom`).
- Then, in the [feature 4 test file](https://github.com/openpharma/clinsight/blob/dev/tests/testthat/test-app_feature_04.R), add a new scenario (scenario 2). 
- In this scenario, temporarily set the environment variable ("CONFIG_PATH") to the correct custom config path location (I think this can be a relative path, but I am not entirely sure). 
- Start a new app instance and take a snapshot (preferably a focused one that does not capture any JSON objects).

As an added bonus, this would also capture the use of a custom config.yml file in an end-to-end test!

@aclark02-arcus would you be willing to try to this? Happy to help here if you have any questions.

NEWS.md
```diff
@@ -1,8 +1,10 @@
 # clinsight (development version)
 
+## Changed
+
 - A double click on a row in the start page now shows a modal with all data that needs review instead of directly going to the first page of the patient even if there is no new data on that page (#216).
 - Removed the requirement for common_forms to have a 'Name' column. In addition, a 'Name' column can be provided in study data tabs, indicating a common name per row, which will improve the query selector items (#207).
-
+- Improved branding with new ClinSight Logo, added favicon, and allowed for study logos, when available
```
LDSamson:
```suggestion
- Improved branding with new ClinSight Logo, added favicon, and allowed for study logos, when available (#214).
```

inst/app/www/logo_mag_for_favicon.png
aclark02-arcus: @LDSamson, yes technically it's not needed anymore, but I thought it I'd retain it just in case we want to use the magnifier somewhere else in the future? Ya never know when it may come in handy.

aclark02-arcus:
> @aclark02-arcus would you be willing to try to this?

Hi @LDSamson, I agree that adding tests are a must, but I think my bandwidth for this PR has been exhausted - i.e. I can't spend more time on it. Too many other competing priorities this week. If you have the time to implement however, I could review the changes? I will add the example study-logo back to the `www/` folder.

LDSamson:
@aclark02-arcus see [here](https://github.com/openpharma/clinsight/pull/214/files/87885c0c6a3509d043215b9024772ac86e49fc72..d7b14586585d8b48b9eedca5531f9504a1f713a7) for the changes

LDSamson:
@aclark02-arcus see the test added in my latest commit which will now fail. Do we need to protect users against this kind of 'misuse' of the assets folder?

LDSamson:
@aclark02-arcus last commit is a way to prevent this potential issue. Easier to understand workflow: just directly point to the study logo in the config file. If it exists, it will be copied in a temporary folder in the temp directory named `clinsight_assets` and then that folder will be made available. 
 
Advantage: Other files in your asset folder will not be made available unintentionally. You could even store the logo in the same folder as study data without any problems, as long as you provide the correct file path.

Disadvantage: A temporary folder will be created and is not automatically removed after session ends. Probably not a big issue.

@aclark02-arcus What do you think? Is this a better approach?

aclark02-arcus (commented):
This looks great @LDSamson! I appreciate the new approach being more explicit.

PR #215, opened 2025-05-22
Example client side logo rendering

LDSamson:
@aclark02-arcus what do you think of something like this?

You can try it out: add a logo folder named "new_assets", place the study logo there, change the golem-config.yml with the path to this folder, and then the app should be able to run with the logo included.

Note that this way, the logo appears almost immediately, instead of only after everything was loaded server side. Since ClinSight can take a long time to load, this might be favorable.

One thing that plays in my mind: The asset folder should not be set to the current working directory. It also should not contain any sensitive information like the entire study data frame. Not sure if we should / need to protect users from doing so. We can discuss.


LDSamson:
Oh and a new logo has arrived! See below, thanks to Patricia. I like it!

image

LDSamson:
@aclark02-arcus another approach for the study logo would be to: - add the full link to the study_logo.png to the golem-config.yml (or .svg or whatever you would like to use). Then create a new temporary assets folder at run time, copy the file in it, and add the resourcepath to this folder. That way we are sure that we only expose this file even if it the path is to the root folder. Thus, it might then become a bit more 'fool-proof'. 
It will, however, create a temporary folder every time you restart the app, although I don't think that would be a big problem.

Thoughts?

LDSamson:
@aclark02-arcus shall we just merge this with your branch? rendering it either server or client side works both I think, but this branch at least fixes the issues you had with assigning an asset folder.

aclark02-arcus (approved):
LGTM! I was able to get this to work using the following code:

### Config

```yml
test:
  study_data: !expr clinsight::clinsightful_data
  meta_data: metadata.rds
  study_asset_path: inst/app/www/assets
```

### R script

```r
devtools::load_all()
if(rlang::is_interactive()){
  raw_data_folder <- app_sys("tests/testthat/fixtures/testapp-raw/data1pt")
  meta_path <- testthat::test_path("fixtures/testapp-raw/metadata_study_example.xlsx")
}
load_and_run_app <- function(){
  # temp_folder <- tempfile(tmpdir = tempdir())
  # dir.create(temp_folder)
  old_golem_config <- Sys.getenv("GOLEM_CONFIG_ACTIVE")
  Sys.setenv("GOLEM_CONFIG_ACTIVE" = "test")
  meta <- get_metadata(meta_path)
  merged_data <- raw_data_folder |> 
    get_raw_csv_data(synch_time = "2024-01-01 00:00:00") |> 
    merge_meta_with_data(meta = meta)
  saveRDS(meta, "metadata.rds")
  run_app(
    onStart = \(){onStop(\(){
      unlink(temp_folder, recursive = TRUE); 
      Sys.setenv("GOLEM_CONFIG_ACTIVE" = old_golem_config)
    })}
  )
}
load_and_run_app()
```

### App

image

PR #216, opened 2025-05-26
Adjust dblclick on start table

LDSamson:
This will change the behavior of a double click on the start page table. This gives a more 'drill down' approach, which gives a better overview of the patient that you are about to review.

clinsight_start_page_double_click

At the same time, I made some of the snapshots smaller so that not literally every snapshot contains the start page JSON object. This will lead to less snapshot changes in the future.


jthompson-arcus (commented):
@LDSamson are you intentionally pointing at the `main` branch?

LDSamson:
> @LDSamson are you intentionally pointing at the `main` branch?

Oops, no that was unintentional

jthompson-arcus (approved):
LGTM

PR #217, opened 2025-05-26
More flexible study status adjustment

LDSamson:
Makes it more flexible to adjust the study status label in ClinSight to any study-specific needs: it skips the standard study status label generation if the variable already exists in the table.

See the adjusted unit test. You can now either add a column named study_status directly to the dataset (the easy, lazy method). Alternatively, you can provide data in long format in which an item_name called 'study_status' is available. 

Lastly, you can provide your own status label, but only if the `study_status` is also available.
Note that the latter functionality is still limited and can just show text in the current implementation.


jthompson-arcus (commented):
Couple of discussion items:

1. I think `study_status` should be renamed as either `patient_status` or `subject_status`. Or maybe something similar. Calling it study status is confusing.
2. I really think this should be a row item and not a new field. I know you wrote it out to be either. In my mind only allowing a row value is just simpler. But maybe I am just unaware of the situation where adding a column makes more sense. Can you elaborate more on that use case?

LDSamson:
> Couple of discussion items:
> 
>     1. I think `study_status` should be renamed as either `patient_status` or `subject_status`. Or maybe something similar. Calling it study status is confusing.
> 
>     2. I really think this should be a row item and not a new field. I know you wrote it out to be either. In my mind only allowing a row value is just simpler. But maybe I am just unaware of the situation where adding a column makes more sense. Can you elaborate more on that use case?

From a developer's perspective, it is easier to just add an additional column if the data is already in long format (as is usually the case with our datasets), rather than recreating a long-format table with study status entries. But I agree that that is cutting corners, the longer format is more consistent with the current data format and does not unnecessarily add the status information to all other forms internally.

What do you mean with that it is easier? Do you mean the code is easier without the second option of adding a column to the data?

LDSamson:
@jthompson-arcus I renamed study_status to subject_status

LDSamson:
@jthompson-arcus I removed the option to provide the `subject_status` by means of a `study_data` column.

R/fct_tables.R
```diff
@@ -183,40 +183,42 @@ create_table.general <- function(
...
   }
-
-  df <- data |> 
-    dplyr::filter(!item_name %in% c("DrugAdminDate", "DrugAdminDose")) |>
+  if ("subject_status" %in% unique(data$item_name)){
+    expected_columns <- c(expected_columns, "subject_status")
+    if ("status_label" %in% unique(data$item_name)){
+      expected_columns <- c(expected_columns, "status_label")
+    }
+  }
```
jthompson-arcus: Feel like something like this could be written more concisely. Maybe something like `expected_columns <- c(expected_columns, intersect(c("subject_status", "status_label"), data$item_name)))`.

Don't feel strongly about it. Guess I'm partial to one-liners.

jthompson-arcus (approved):
@LDSamson LGTM. I did add one comment but I'm going to mark this as approved. If you want to do the edit I suggested, I will just re-approve. Otherwise feel free to mark that convo as resolved and merge.

R/fct_tables.R
LDSamson: Thanks. That's a bit more concise indeed. I will add it

R/fct_tables.R
```diff
@@ -185,7 +185,7 @@ create_table.general <- function(
     data <-  data.frame(matrix(ncol = length(df_names))) |> 
       setNames(df_names)
   }
-  df <- with(data, data[!item_name %in% c("DrugAdminDate", "DrugAdminDose"),]) |>
+  df <- data[!data[[name_column]] %in% c("DrugAdminDate", "DrugAdminDose"),] |>
```
LDSamson, line 188: @jthompson-arcus honestly, I don't remember the reason for filtering these out here; maybe it can even be removed.

EDIT: the reason was that, in a very old version, these columns could cause duplicates. But that is already resolved now since `create_table.default` collapses multiple values into one string if needed:
image.

I guess we can leave it in, I do not have a strong opinion about it.

aclark02-arcus: Let's leave it!

PR #219, opened 2025-06-11
Add unit tests

LDSamson:
Nothing special, this just tests some edge cases that were previously not yet covered by unit tests. Hardly any changes are made in the actual code base.

It also places the function db_update_version to the dev folder since we do not plan to support this function in the current form; it is only meant to update from the old database format that hopefully no one else but us used, and it is not needed for setting up a new study with the application.


tests/testthat/test-fct_SQLite.R
```diff
@@ -386,6 +423,26 @@ describe(
...
+  it("returns an empty string if the version is not available", {
+    temp_path <- withr::local_tempfile(fileext = ".sqlite") 
+    con <- get_db_connection(temp_path)
+    expect_equal(db_get_version(temp_path), "")
+  })
+  it("errors if database is uavailable", {
+    expect_error(db_get_version("fake_path/db.sqlite"))
+  })
+})
+
```
aclark02-arcus, line 445: Do we need tests for functions that aren't in `R/`?

R/fct_appdata.R
```diff
@@ -284,14 +284,10 @@ get_appdata <-  function(
...
     item_group_x <- unique(x$item_group)
-    if(length(item_group_x) != 1) stop(
-      "item_group consists of multipe elements which is not allowed: ", 
-      item_group_x
-    )
     form_type_x <- unique(with(var_levels, form_type[item_group == item_group_x]))
     if(length(form_type_x) != 1) stop(
       "form_type consists of multipe elements which is not allowed: ", 
-      form_type_x
+      paste0(form_type_x, collapse = ", ")
```
aclark02-arcus, line 290: ha, we already made this change on our side!

aclark02-arcus (approved):

tests/testthat/test-fct_SQLite.R
LDSamson: I think these functions are available in `R/`, right?

aclark02-arcus: Oh sorry, I was confusing `get_db_version()` with `update_db_version()` which now lives in `dev/`

PR #225, opened 2025-06-17
Fix event id order estimation

LDSamson:
When event label order needs to be estimated (in rare cases where the event id order is not consistent in all subjects based on the date), the arranging was by mistake internally done by event_id instead of the estimated order. This will become apparent if for example event_id contains something like "Vis1" "Vis2" "Vis10" in which case the ordering of this character vector will be incorrect with the old behavior.

This PR fixes this behavior.


R/fct_data_customize_events.R
```diff
@@ -186,7 +186,7 @@ add_events_to_data <- function(
         dplyr::mutate(order_occurrence = dplyr::n(), .by = derived_order) |> 
         dplyr::slice_min(order_occurrence, by = event_id, with_ties = FALSE) |> 
         dplyr::select(-order_occurrence) |> 
-        dplyr::arrange(event_id, derived_order) |> 
+        dplyr::arrange(derived_order, event_id) |> 
         dplyr::mutate(derived_order = dplyr::row_number())
```
LDSamson, line 190: @aclark02-arcus this is basically all that was needed to fix the issue

R/fct_data_customize_events.R
```diff
@@ -215,7 +215,7 @@ add_events_to_data <- function(
         paste0(event_label_custom, vis_number),
         event_label_custom
       ),
-      event_label_custom = factor(event_label_custom, levels = unique(event_label_custom))
+      event_label_custom = factor(event_label_custom, levels = unique(na.omit(event_label_custom)))
     )
```
LDSamson, line 219: @aclark02-arcus removing missing values just because it is a bit cleaner

aclark02-arcus (approved):
🔥

PR #228, opened 2025-06-25
Improve continuous data S3 table class

LDSamson:
Small change. Harmonizes the 'Reviewed' label in the interactive table, and adds event date to the `create_table.continuous()` S3 class.

grafik

Indeed, it is captured in too many snapshots, too many files are changed. But I do not have time to make the snapshots more targeted right now, and the change in the snapshots is almost everywhere the same, so hopefully it is not a big issue for the review.


jthompson-arcus (approved):
LGTM

PR #230, opened 2025-07-09
Update test utils

LDSamson:
Just an update of a few developer functions. This has no impact on ClinSight because the functions are not directly used, only to create the vignette.

Only two changes:

1) The parameter `include_skipped` is now properly propagated so that it can be changed in the exported function
2) more information will now be printed/shown if errors and/or warnings occur or if files were skipped


aclark02-arcus (approved):
LGTM

PR #231, opened 2025-07-09
Update package versions

LDSamson:
Update R version, and update package versions to the latest snapshot date.

Changed:
-  `bslib::navbar_options` must now be used to set some navbar options
    - Therefore, I added bslib >= v0.9.0 as dependency in the `DESCRIPTION`. Let me know if you think that is not a good idea.
- `is.ggplot` is renamed to `is_ggplot` (only affects tests)
- Many snapshot updates because of updated packages


LDSamson:
@aclark02-arcus I updated the snap for a final time, mainly to include the latest `httr2` version that contains significant changes we would like to use internally.

Do you have any update on when you can look at this and the other open PR? If I can help with anything, or if you would like to discuss something, let me know!

aclark02-arcus:
@LDSamson, I don't think I have the bandwidth to look at this and rigorously test this right now. I have way too many irons in the fire at the moment, so if there is someone else available to review, it may be wise to ask them... 👀 

My sincerest apologies!

LDSamson:
> @LDSamson, I don't think I have the bandwidth to look at this and rigorously test this right now. I have way too many irons in the fire at the moment, so if there is someone else available to review, it may be wise to ask them... 👀
> 
> My sincerest apologies!

No problem, I understand! Thanks for letting me know!

vanguyen317 (approved):

PR #232, opened 2025-07-10
Fix namespacing issues of exported `key_columns` object

### LINKED ISSUE #233: Exported `key_columns` object results in multiple obscure namespacing issues
@jthompson-arcus and @aclark02-arcus: I think we should remove the export from the `key_columns` object. This export leads to unexpected namespacing errors because of package data is loaded lazily (LazyData is set to TRUE in the description). In short, in contrast to internal data objects, exported data objects are apparently not intended for use within package functions. 

This leads to multiple issues. For example, you will not be able to call `clinsight::db_update()` without loading `ClinSight` first.
Also clinsight itself will not work (calling clinsight::run_app() without loading the package first). 
Worse than this, if you load `ClinSight`, but then define a key_columns variable globally, this variable will then clash with `clinsight::key_columns` and be used within clinsight::db_update() instead. 

There are work arounds, but they all make it more complicated to use `key_columns`, or they circumvent best practices of R package development. In my opinion, we should just be able to safely use this variable without the need of, for example, calling clinsight::key_columns every time, and it is not worth the issue to hack our way out of it for such a small character vector.

LDSamson:
Fixes #233 by unexporting the key_columns object.  *EDIT*: decided to use a different form of export, as discussed [here](https://github.com/openpharma/clinsight/pull/235): 

In addition, it removes some of the hard coded `key_colums` strings, and harmonizes the key column arguments within the internal functions. Documentation of `key_columns` can now be found in the Metadata vignette


LDSamson:
Reconsidered export of the vector, see discussion here: https://github.com/openpharma/clinsight/pull/235.

Now exporting the vector by declaring it in `R/data` and using an `@export` tag.

LDSamson:
@jthompson-arcus should be ready now

R/sysdata.rda
jthompson-arcus, line 1: @LDSamson why is this changing?

jthompson-arcus (commented):
This looks good to me. Just had the one observation that for some reason a change in `R/sysdata.rda` is being tracked. I can't tell what's different about it. No changes in `data-raw/internal_data.R` either.

R/sysdata.rda
LDSamson: Strange. I just ran a quick comparison but found no differences for any of the objects inside the file: 

```
old_env <- new.env()
new_env <- new.env()
data_old <- load(file = "sysdata_old.rda", envir = old_env)
data_new <- load(file = "R/sysdata.rda", envir = new_env)

for (n in data_old) {
  print(waldo::compare(get(n, old_env), get(n, new_env)))
}
# ✔ No differences
# ✔ No differences
# ✔ No differences
# ✔ No differences
# ✔ No differences
# ✔ No differences
# ✔ No differences
# ✔ No differences
```

All I can think of is that it might be created with the latest R version instead of the one mentioned in the lockfile and that therefore for example the compression is slightly different.

@jthompson-arcus I reverted to the old file to be sure

jthompson-arcus (approved):
LGTM

PR #234, opened 2025-07-11
Update test

LDSamson:
I noticed that the module test for `mod_review_form_tbl` was referencing the wrong module. This will fix it. In addition, I've set up two very basic tests to ensure the module works in isolation testing basic functionality.


tests/testthat/test-mod_review_form_tbl.R
```diff
@@ -35,102 +35,116 @@ describe(
...
+          ae_rev_status <- merged_form_data() |> 
+            dplyr::filter(
+              subject_id == active_subject(), 
+              grepl("Allergic Reaction", Name)
+            ) |> 
+            dplyr::pull(row_review_status)
+          ae_rev_status <- ae_rev_status[[1]]
+          session$setInputs(
+            table_review_selection = data.frame(
+              id = ae_rev_status$id,
```
jthompson-arcus:
```suggestion
              id = ae_rev_status$ids, 
```
The named value is actually `ids`. R is doing a partial match here.

tests/testthat/test-mod_review_form_tbl.R
```diff
@@ -35,102 +35,116 @@ describe(
...
+          session$setInputs(
+            table_review_selection = data.frame(
+              id = ae_rev_status$id, 
+              row_id = ae_rev_status$row_id, 
+              reviewed = "TRUE"
+            )
+          )
+          expect_equal(
+            session$userData$pending_review_records[[form]],
+            data.frame(id = ae_rev_status$id, reviewed = "TRUE")
```
jthompson-arcus:
```suggestion
            data.frame(id = ae_rev_status$ids, reviewed = "TRUE")
```

tests/testthat/test-mod_review_form_tbl.R
```diff
@@ -35,102 +35,116 @@ describe(
...
+              subject_id == active_subject(), 
+              grepl("Allergic Reaction", Name)
+            ) |> 
+            dplyr::pull(row_review_status)
+          ae_rev_status <- ae_rev_status[[1]]
+          session$setInputs(
+            table_review_selection = data.frame(
+              id = ae_rev_status$id, 
+              row_id = ae_rev_status$row_id, 
+              reviewed = "TRUE"
```
jthompson-arcus:
```suggestion
              reviewed = "Yes"
```
The input handler returns "Yes", is there a reason to use "TRUE"?

tests/testthat/test-mod_review_form_tbl.R
```diff
@@ -35,102 +35,116 @@ describe(
...
-    #           expect_true(inherits(output[["review_form_tbl-table"]], "json"))
-    #         })
-    #       })
+    it(
+      "Scenario 1 - Select row for review. Given a test [Adverse events] data set,
+        and the active subject_id set to ID 'DEU_02_482',
+        and the subject having an adverse event [Allergic Reaction] that has not yet been reviewed,
+        and setting the rows of the event [Allergic Reaction] to be selected in the form table,
+        I expect that [pending_review_records] gets updated with the approriate row numbers,
+        and that the output table is a valid JSON object.",
```
jthompson-arcus: I don't know if I'm being overly picky here, but this first scenario test bothers me a little. I think it's because the test does not actually "set the rows to be selected". I'm not sure what the solution is but thought we should discuss.

jthompson-arcus (changes requested):
@LDSamson few comments below

tests/testthat/test-mod_review_form_tbl.R
LDSamson: @jthompson-arcus Would 'Mark rows as selected for review' be a better description here?

LDSamson:
@jthompson-arcus anything else you'd like to check or change here before merging?

jthompson-arcus (approved):
LGTM

PR #235, opened 2025-07-11
Alternative exporting of `key_columns`

jthompson-arcus:
@LDSamson have you considered this approach?

Related to #233


LDSamson:
@jthompson-arcus  Yes, I considered it, but it goes against best practices of the authors of testthat/Roxygen2. See https://r-pkgs.org/data.html#sec-documenting-data, last sentence: 

> Never @export a data set.

I feel like if the authors of r-pkgs have such a strong opinion against this, it is clearly not something that they expect people to do, and I am a bit afraid that this will lead to issues in the future (for example, maybe they will try to prevent people using the export tag this way to discourage the not intended usage?)

And then, when you remove the export tag, it is just the same as just using the internal data structure as before

jthompson-arcus:
I definitely don't read that paragraph the same way. You would not want to use an `@export` if your dataset is stored in `data/` as an RDA file.

jthompson-arcus:
I would also like to point out that it doesn't even register as a NOTE in the R CMD check. Also, it is not hard to find examples within the tidyverse itself where objects like this are exported.

For example, check out `ggplot2::.stroke`.

LDSamson:
Interesting. I searched for examples but could not find them. Since you said it was common: do you have more examples were this was used apart from `ggplot2::.stroke`/`ggplot2::.pt`?
 
It is at least not a clearly documented/recommended way of dealing with data though. See 'good practice' section from the base R authors here: https://www.rdocumentation.org/packages/utils/versions/3.6.2/topics/data and r-pkgs.org: https://r-pkgs.org/data.html#sec-data-data

But I guess we could do it that way if it is common practice, the risk is not too high, and it would help you

jthompson-arcus:
> Interesting. I searched for examples but could not find them. Since you said it was common: do you have more examples were this was used apart from `ggplot2::.stroke`/`ggplot2::.pt`?
> 
> It is at least not a clearly documented/recommended way of dealing with data though. See 'good practice' section from the base R authors here: https://www.rdocumentation.org/packages/utils/versions/3.6.2/topics/data and r-pkgs.org: https://r-pkgs.org/data.html#sec-data-data
> 
> But I guess we could do it that way if it is common practice, the risk is not too high, and it would help you

Few examples from the tidyverse:

### ggplot2

- `.pt`: https://github.com/tidyverse/ggplot2/blob/f666daff39234dd30541b47e688fd93e74502cd8/R/geom-.R#L512
- `.stroke`: https://github.com/tidyverse/ggplot2/blob/f666daff39234dd30541b47e688fd93e74502cd8/R/geom-.R#L515

### rlang

- `.data`: https://github.com/r-lib/rlang/blob/7371bac25ffe9b91097e0dad424cec5b2ee3c288/R/eval-tidy.R#L210
- `.env`: https://github.com/r-lib/rlang/blob/7371bac25ffe9b91097e0dad424cec5b2ee3c288/R/eval-tidy.R#L215

### lubridate

- `NA_Date_`: https://github.com/tidyverse/lubridate/blob/c073b81fac869d9ed4771bfa4784e1b1aae12d1a/R/Dates.r#L28
- `origin`: https://github.com/tidyverse/lubridate/blob/c073b81fac869d9ed4771bfa4784e1b1aae12d1a/R/instants.r#L64

And there are many, many more examples. What you typically don't see are data frames done this way. We are not passing a data frame in this way but a small character vector.

Additionally, I still hold to the position that the documentation you are referencing is specifically related to data attached to a package via `data(...)`, not exported constants and environments.

LDSamson:
Closing this since it is solved in #232

PR #236, opened 2025-07-16
Update test suggestion

jthompson-arcus:
@LDSamson how do you feel about something like this. We are at least using the handler the app uses now. Still feel a little bit hokey, but I think I like it better.


LDSamson:
Looks good to me. I will merge it.

I would suggest to also include pending_review_records as as a test value export (`shiny::exportTestValues()`) so that we can check if the ids returned in the module are as expected. I will add that as a next step in the PR that I created, after merging this one.

LDSamson (approved):

PR #237, opened 2025-07-17
Acknowledge `edit_date_time` values during `db_update()`

aclark02-arcus:
PR to `update_review_data()` that only consider updating records with a new `edit_date_time` that are more recent than the "old" `edit_date_time` to determine if the data should be considered new/updated.

Why do we need this? We're going to be changing a bunch of `edit_date_time` values for a study that has a production app deployed. Right now, if we attempt to update the db with the new `merged_data` object, it marks all those records as needing review. However, if some records get a new `edit_date_time` from a prior date, we want it to retain the previous review completed.

Originally authored by @jthompson-arcus! I can't take any credit. We covet your thoughts @LDSamson!

### DIFF
--- a/R/mod_review_data_fct_helpers.R
+++ b/R/mod_review_data_fct_helpers.R
@@ -91,6 +91,10 @@ update_review_data <- function(
     # including event_date in join by should work, even for Adverse events, since 
     # if event_date changes, the edit date-time also changes, and thus all rows 
     # will be taken into consideration.
+    dplyr::left_join(review_df, by = key_cols, suffix = c("", ".old")) |>
+    dplyr::filter(is.na(.data[[paste0(edit_time_var, ".old")]]) | 
+                    .data[[edit_time_var]] > .data[[paste0(edit_time_var, ".old")]]) |>
+    dplyr::select(dplyr::all_of(names(latest_review_data))) |>
     dplyr::full_join(review_df, by = names(latest_review_data)) |> 
     add_missing_columns(c("timestamp", "reviewed", "comment", "status"))
   


LDSamson (commented):
Hi Aaron,

initial thoughts: 
- it feels like quite a rare and strange case when edit date times in your updated data are older than in the old database. Do you expect this to happen on a regular basis? Or is this meant as a one-time change to address issues in the current database?
- Wouldn't it warrant at least a warning if something like this occurs since it might also happen when there are synchronization issues?
- first impression is that it would not interfere with databases where the edit_date_time is correctly implemented, so that is a good thing. I will check more in detail next week. 
- Is it needed to run it every time an app starts? If so, would this slow down the app more with big datasets? Some quick tests here would maybe be nice because the app is already too slow in start up. Maybe it's not an issue at all but I am not sure at the moment

Apart from the points above, some more practical things: it would be great to capture this behavior in a test, and maybe add a comment in the function itself why this additional filter is actually needed.

What would be your desired time frame to implement this? Is it needed for release v0.3.0, or can it also be in a follow-up PR in the dev branch?

don't have too much time today to look further into it, but I am happy to discuss further of course!

aclark02-arcus:
> it feels like quite a rare and strange case when edit date times in your updated data are older than in the old database. Do you expect this to happen on a regular basis? Or is this meant as a one-time change to address "issues" in the current database?

@LDSamson We suspect it's primarily a one one time change, but like you said, it doesn't appear to hurt anything. In fact, it could be seen as a nice "safety feature"... to only bold records when their edit date times are new, just in case there is some issues with timestamps in the EDC.

> Wouldn't it warrant at least a warning if something like this occurs since it might also happen when there are synchronization issues?

Yes, agreed. We can work on adding that.

> Is it needed to run it every time an app starts?

No, from what I can see, `update_review_data()` is only run in `db_update()`... Oh wow, I just looked at where `app_prod` is used to run `db_update()` in `run_app()`. Looks like we have `app_prod: no` in the golem config but have `options( "golem.app.prod" = TRUE)` in our Posit Connect deployment. 🤣 

We don't run (and don't want to run) `db_update()` in `run_app()`. Maybe that's why we did that. `app_prod: yes` would make app start up unbearable for us, lol. We update the db in a different process that is scheduled to kick off every morning.

> What would be your desired time frame to implement this? Is it needed for release v0.3.0, or can it also be in a follow-up PR in the dev branch?

Follow-up PR to the `dev` branch would be fine. We'll likely implement it internally first because we need it ASAP.

aclark02-arcus:
@jthompson-arcus, do you want to bring this PR up to date before we close it? Just for completeness in case we ever need to refer back to it?

LDSamson:
As discussed during dev meeting on 14 Oct 2025, can be closed

PR #241, opened 2025-08-12
Add "hidden" download handler on custom DataTable

### LINKED ISSUE #240: Excel download lists are incomplete
When using the download button, the downloaded Excel lists are incomplete. See screenshots below. Probably because of the `deferRender` option from the `Scroller` `datatables` extension that we are using. 

@jthompson-arcus @aclark02-arcus do you know how we can fix this?

Image

Image

jthompson-arcus:
Addresses #240


jthompson-arcus:
@LDSamson & @aclark02-arcus there is still a lot of work to be done to implement a server-side download but what do you think of this approach?

jthompson-arcus:
@LDSamson I look further into this and I believe this is the best solution without creating something custom.

jthompson-arcus:
@LDSamson can you remind me what changes you wanted made to the output here?

R/mod_review_form_tbl.R
```diff
@@ -221,6 +224,31 @@ mod_review_form_tbl_server <- function(
...
+      filename = function() {
+        export_label = paste(
+          ifelse(identical(title, "Serious Adverse Events"), "SAEs", simplify_string(form)), 
+          ifelse(show_all(), "all_patients", active_subject()), 
+          sep = "."
+        )
+        paste("clinsight", export_label, "csv", sep = ".")
+      },
+      content = function(file) {
+        readr::write_csv(
```
LDSamson, line 238:
```suggestion
        readr::write_delim(
```

R/mod_review_form_tbl.R
```diff
@@ -221,6 +224,31 @@ mod_review_form_tbl_server <- function(
...
+          table_data() |> 
+            subset(show_all() | subject_id == active_subject()) |> 
+            dplyr::select(-row_review_status) |> 
+            dplyr::rename(dplyr::any_of(table_names)) |> 
+            dplyr::mutate(dplyr::across(
+              dplyr::where(is.character),
+              \(x) gsub("<b>|</b>", "", x)
+            )), 
+          file,
+          na = ""
```
LDSamson, line 248:
```suggestion
          delim = ";",
          na = ""
```

LDSamson (commented):
Looks good to me. 

Two things: 

1. Would changing the download to a file with a semicolon separator be okay for you? See my suggestion. It makes it much easier for us Europeans to open the file in Excel without additional steps, because Excel settings in most European countries expect a comma as decimal separator. Excel quirks..
2. I will check if we can add a basic test to ensure the downloadable file is as expected, with the correct number of rows. Not sure how to capture the JS click, but we can add at least a test that the shiny part delivers the correct expected table. I will add this in a separate branch based on yours. Should hopefully work with a simple `testServer` test.

LDSamson (approved):
LGTM. As discussed,  we can add option for a different delimiter in a later PR, by making it configurable with an option.

PR #243, opened 2025-09-08
Improve `add_row_review_status()`

jthompson-arcus:
We have found that `ifelse()` is way more efficient than `dplyr::case_when()` (by order of 2x as fast). This becomes more critical as data increases in size. We have tables where the current process can take upwards of 3 seconds to calculate. This small change moves that number closer to 1 second.


LDSamson (approved):
LGTM

Issue #223, opened 2025-06-17
Minimize app startup time


aclark02-arcus:
@jthompson-arcus, I recall chatting with you once about reducing the amount of data that's processed at startup. I think you suggested that maybe that the objects at the top of app_server.R could be generated before calling run_app(), correct? Just thought I'd open an issue for it to keep that thought moving forward.

LDSamson:
Do you mean before calling `shinyApp` within the `run_app` function? We can try that. 

Pulling object generation outside of `run_app` is a bigger design change but can also be considered. 

We should first find the largest bottlenecks I think. 

Related: 
- https://github.com/openpharma/clinsight/discussions/211
- https://github.com/openpharma/clinsight/discussions/210

aclark02-arcus:
I want to give this issue a `+1` - we went live for a larger study last week and the app is significantly lethargic, and not only at app start, but **especially at app start.** 

Perhaps a different issue should be opened for this, but even navigating through the app is slow. For example, clicking a new tab that hasn't been opened yet, clicking "Save" on records, etc.

LDSamson:
@aclark02-arcus can you perform some profiling and check what is the largest bottleneck?
How much data is in the app for the study that you are mentioning?
And how many items do you have per form? 
Is it only slow in the figure view or also in the table view?

aclark02-arcus:
<img width="146" height="50" alt="Image" src="https://github.com/user-attachments/assets/4e67f934-0a1a-4386-b900-65a96e5105ca" />

703,118 items total.
Largest form is AEs @ 133,142 items.

The slowness is in the tables.

LDSamson:
@aclark02-arcus what makes your user database so big? Do you think that is the biggest culprit?
Maybe it is the pivoting in `create_table` that causes the issue? Can you check how long it takes to call `create_table` on `appdata[["Adverse events"]]`?

Did you already update to the latest clinsight version? Using the latest version could have some performance gains due to updated R system dependencies in version 4.5.x (although I don't know how much)

LDSamson:
Simple script to test big datasets using dummy data: 

```
library(tidyr)
library(dplyr)
library(clinsight)
##
duplications <- 50
##
df <- clinsightful_data |> 
  select(-site_code) |> 
  tidyr::separate_wider_delim(subject_id, "_", names = c("country", "site", "subject"))
big_data <- lapply(
  seq_len(duplications), 
  \(x) {
    df |> 
    mutate(
[...]

PR #245, opened 2025-09-10
Preprocess heavy computations before `run_app()`

aclark02-arcus:
Significant work towards #223.

This PR (in addition to #243) reduced app startup time from 2m 40s down to 41s - 70s. Note: these time stats were measured on a larger study with ~600-700 patients. Also, the amount of time to simply review a record went down from 27s down to 12s from these two PRs.

Obviously a lot of changes here. Feel free to take your time & suggest alternatives, but would love some initial thoughts if you have any. For now, I just commented out some old code, but when this PR is ready, I'm happy to remove it altogether.

For some reason, I had to add a couple lines to `dev/run_dev.R/load_and_run()` and `test_clinsight()` to build the db first... not sure why. Maybe you know?


LDSamson:
Hi @aclark02-arcus,

The application surely need improvement in startup time, so I think this is indeed the correct way forward. Good news that these changes helped to reduce the start up time for you!

We probably need to implement this in something like as two-step process. As discussed, maybe you can merge it first in your `devex` branch, and then we can check how we can validate the changes properly before merging into `dev`.

For validation according to GAMP V guidelines, we should identify potential risks of this adjustment, provide a risk classification, and then try to mitigate especially high priority risks.

For example, in this case, we are changing the core files in memory. Since we would now rely on more files, they might go out of synch, which could be a risk. We could mitigate this by:
- Reducing the number of files that we need to use (maybe put some more files in lists? Is that possible)
   - Can we maybe even remove study_data and just rely on app_data? Not using study_data will also help with memory usage of the app
   - Do we still need to rely on the app_tables object, or is this only needed for creating one of the other overview tables? 
- add a time stamp as attribute to each file, so that it is clear that they were created at the same moment, and verify that the time stamp matches.
- ...

Another risk is that we would like to update to the new system without much problems. Is this possible or would it need a major update? Probably some documentation update on how to update would also be good.

Let me know when you are available, then we can discuss this!

aclark02-arcus:
Hi @LDSamson, thanks for looking at this. Just a quick note: to my knowledge, we've abandoned `devex` altogether, but we could bring it back if you wish. We'd probably "restart" by basing it off the current `dev` branch. Is there a reason you want to use `devex` and instead of `dev`? Seems like we could just work out all of those GAMP considerations here (in this PR). What do you think?

Just so you know, we've already committed these changes into our internal `clinsight` branch. But we'd welcome anything new that may come up during PR review.

LDSamson:
Ah, good to know. I thought you would like to move forward quickly using such a devex branch.

Main reason I was a bit hesitant for using `dev` is that I would like to be able to support current versions that are running in production, with small updates/bug fixes if needed, without having to make big changes to the setup in studies that are already running. If this PR would still allow the old setup, and if everything is tested/validated properly before merging, then merging into `dev` would also be fine.

LDSamson (changes requested):
See previous comment. The risk of this change seems manageable, just good to have a few adge cases covered :) If you want to discuss let me know!

aclark02-arcus:
Now that #251 has merged, `app_tables` doesn't really exist, so I'll change the name of this PR. Regardless, this PR is still worthwhile for it's pre-processing of the data before launching the app.  Commit a8fbcb1 got rid of `app_tables`, but added `timeline_data` to the pre-processing step to save lots of time within the AE tab. It's now being saved as both an .rds & .parquet file.

data-raw/clinsightful_data.R
```diff
@@ -1,9 +1,48 @@
...
+#     create_table(cs_app_data[[x]], expected_columns = names(cs_app_vars$items[[x]]))
+#   })
+
+# Build a 'available_data'
+cs_available_data <- get_available_data(
+  data = cs_app_data,
+  # tables = cs_app_tables,            # outdated arg
+  # all_forms = cs_app_vars$all_forms, # outdated arg
+  form_repeat_name = with(
+    meta[["table_names"]],
```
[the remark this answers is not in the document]
aclark02-arcus: resolved with commit [cfe7130](https://github.com/openpharma/clinsight/pull/245/commits/cfe7130ce75f5a255ad89b0c7f28c7ebd4a3afab)

man/metadata.Rd
```diff
@@ -7,6 +7,66 @@
...
+#>   ..$ item_name  : chr [1:220] "AE Number" "AE Name" "AESI" "AE start date" ...
+#>   ..$ item_type  : chr [1:220] "other" "other" "other" "other" ...
+#>   ..$ item_group : chr [1:220] "Adverse events" "Adverse events" "Adverse events" "Adverse events" ...
+#>   ..$ unit       : chr [1:220] NA NA NA NA ...
+#>   ..$ lower_limit: chr [1:220] NA NA NA NA ...
+#>   ..$ upper_limit: chr [1:220] NA NA NA NA ...
+}\if{html}{\out{</div>}}
+}
+
+\subsection{\code{metadata}: A list with 9 objects.}{
```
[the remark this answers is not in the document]
aclark02-arcus: resolved with [b5d631f](https://github.com/openpharma/clinsight/pull/245/commits/b5d631f03c29bb50aca94d5e109d71a15e565a85)

man/cs_timeline_data.Rd
```diff
@@ -0,0 +1,38 @@
+% Generated by roxygen2: do not edit by hand
+% Please edit documentation in R/data.R
+\docType{data}
+\name{cs_timeline_data}
+\alias{cs_timeline_data}
+\title{Application Timeline Data}
+\format{
+\subsection{\code{cs_timeline_data}: data.frame formatted for comsumption by}{
```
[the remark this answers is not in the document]
aclark02-arcus: Resolved with [7f84adc](https://github.com/openpharma/clinsight/pull/245/commits/7f84adc426ee444c588fb19e3cd0a52c3a2fee5b)

aclark02-arcus:
@LDSamson, pinging you for a review to make sure you are content with the recent changes.

R/run_app.R
```diff
@@ -125,6 +182,10 @@ run_app <- function(
     ),
     golem_opts = list(
       meta = meta,
+      app_data = app_data,
+      app_vars = app_vars,
+      app_tables = app_tables,
+      available_data = available_data,
       data = data,
```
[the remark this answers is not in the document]
aclark02-arcus: Thanks! Commented this out.

dev/app.R
```diff
@@ -15,11 +15,58 @@ merged_data <- merge_meta_with_data(
...
+  "timeline_data")
+purrr::walk(save_objs, function(x){
+  rds_file <- file.path(data_folder, paste0(x, ".rds"))
+  saveRDS(get(x), rds_file)
+  if(inherits(get(x), "data.frame")) {
+    pq_file <- file.path(data_folder, paste0(x, ".parquet"))
+    arrow::write_parquet(get(x), pq_file)
+  }
+})
+
```
LDSamson, line 69: This looks quite difficult to start the app.. you must remember to create all these objects, name them correctly, feed them into the app.. can't we simplify this process more?

R/run_app.R
```diff
@@ -39,23 +44,51 @@ run_app <- function(
...
+  if(is.character(app_data)){
+    if(!file.exists(app_data)) stop(paste0("Cannot find '", app_data, "'."))
+    if(tolower(tools::file_ext(app_data)) != "rds"){
+      stop("Invalid 'app_data' format. Expecting a file .rds format")
     }
-    data <- readRDS(data)
+    app_data <- readRDS(app_data)
   } 
-  stopifnot("Expecting study data to be in data frame format." = is.data.frame(data))
+  stopifnot("Expecting 'app_data' to be in list format." = inherits(app_data, "list"))
```
LDSamson, line 79: I might be wrong here, but I have a feeling that the improvement in loading time does not come so much from using parquet files. Reason I think that is that for the largest object, the study data here, no parquet file is used. Can't that be changed?

R/run_app.R
LDSamson, line 156: > For some reason, I had to add a couple lines to dev/run_dev.R/load_and_run() and test_clinsight() to build the db first... not sure why. Maybe you know?

It's because the database is created in this step here when it does not exist. But this will error now, because the `data` object was deleted by you (see line 27).

R/fct_test_clinsight.R
```diff
@@ -32,17 +32,67 @@ test_clinsight <- function(
...
+    "timeline_data",
+    "available_data")
+  purrr::walk(save_objs, function(x){
+    rds_file <- file.path(temp_folder, paste0(x, ".rds"))
+    saveRDS(get(x), rds_file)
+    if(inherits(get(x), "data.frame")) {
+      pq_file <- file.path(temp_folder, paste0(x, ".parquet"))
+      arrow::write_parquet(get(x), pq_file)
+    }
+  })
```
LDSamson, line 95: I think adding parquet files here will make it significantly slower to test clinsight with this function because it is fairly slow to write parquet files. Can this be changed/removed here?

R/fct_test_clinsight.R
```diff
@@ -32,17 +32,67 @@ test_clinsight <- function(
...
+  )
+  save_objs <- c(
+    "clinsight_data",
+    "meta_data",
+    "app_data",
+    "app_vars",
+    # "app_tables",
+    "timeline_data",
+    "available_data")
+  purrr::walk(save_objs, function(x){
```
LDSamson, line 88: Keep in mind that `purrr` is currently only in `suggests` in the description. We should either use base R or move the package to `Imports`.

LDSamson (changes requested):
Hi @aclark02-arcus, thank you for your work on this.

See also the few detailed comments I made. 

In general, the setup of a study feels a bit messy now because we need to create so many objects with the correct naming before we can even start clinsight. It would be great if that can be addressed. 

Suggestions (open for debate of course) would be to: 

1. create a list object (e.g. `clinsight_data`) with the following objects:
- appdata (list created with `get_appdata`)
- timeline_data (data frame with timeline_data)
- available_data (data frame with available_data)
- app_vars (list with app variables/metadata)

2. create a helper function to create this list object, with the `merged_data` object and the `metadata` object
3. create a helper function to save `clinsight_data` with a boolean option `use_parquet` (by default FALSE to be compatible).
    - eithers saves it as a single .rds file, or as different parquet files
5. create a helper function to load `clinsight_data`, with the same option `use_parquet`

This way, we only have to feed `clinsight_data` to `run_app` and not so many different objects, whil still using data pre-computations.

Furthermore, if we are using parquet files, shouldn't we also save all the data frames in `appdata` as a parquet file? These are the largest objects by far. 

Looking forward to hearing your thoughts on this

R/run_app.R
aclark02-arcus: I can't remember what the `run_app()` time savings was by switching from `.rds` to `.parquet` for data.frame objects but I remember it was "pretty decent."  Note: switching to parquet decreased the data build time by 18% for a large study... as saving large data.frames as .rds is slow. BUT the best advantage here was the file storage efficiencies when parquet compressed the daily build size from 10GB to 1GB... and that was without any changes to `app_data`.

But I agree with what you are saying here. I didn't want to mess with changing the `app_data` object since it is a list & parquet can't help with that unless you break apart the list.

aclark02-arcus:
> Looking forward to hearing your thoughts on this

Yes, I'm not opposed to a change like that. It is fairly "messy" as you say. I won't be able to work on it in the near future however with my OOO in Jan though.

R/run_app.R
LDSamson: > But I agree with what you are saying here. I didn't want to mess with changing the app_data object since it is a list & parquet can't help with that unless you break apart the list.

Can't you just place the data frames within `app_data` in a separate parquet file in a folder (folder named `appdata`), and then read all files in this folder with parquet using a lapply loop? Or is that not possible with `parquet` files? Note that you also must assign the class and the form name again but that should not be too hard.

LDSamson: > I can't remember what the `run_app()` time savings was by switching from `.rds` to `.parquet` for data.frame objects but I remember it was "pretty decent." Note: switching to parquet decreased the data build time by 18% for a large study... as saving large data.frames as .rds is slow. BUT the best advantage here was the file storage efficiencies when parquet compressed the daily build size from 10GB to 1GB... and that was without any changes to `app_data`.

Interesting, thanks for sharing. So the change was mostly in building datasets, as in the `merged_data`  object?
I thought the file size was bigger with parquet files because it focuses more on build speed.
My point was not that is saves time to use `parquet` files but that these files are not used so much here, only for `timeline_data` and `available_data`.

PR #248, opened 2025-10-28
Also fix long-format Multiple Choice variables that end with a number

### LINKED ISSUE #247: Multiple choice questions are adjusted incorrectly during data pre-processing if variable ends with a number
When setting up a study, multiple choice questions can be added in a long-format and these will be adjusted with the internal function `fix_multiple_choice_vars()`. However, this function does not work as intended if the original variable ends with a number. 

Thus, if the original variable is called `MC_VAR1` and the data looks like this:

|  var  | item_value | 
| ----- | ------------ |
| MC_VAR11 | option1 |
| MC_VAR12 | option2 |
| MC_VAR13 | option3 |

Then, after using the adjustment function, the outcome is: 
|  var  | item_value | 
| ----- | ------------ |
| MC_VAR | option1; option2; option3 |

While it is expected to be:
|  var  | item_value | 
| ----- | ------------ |
| MC_VAR1 | option1; option2; option3 |

This can be an issue during study-specific setup. The issue is not so visible if you are unaware of it, therefore I label it as medium-risk priority.

LDSamson:
Fixes #247.


jthompson-arcus (approved):
Everything seems to be working as intended.

Issue #110, opened 2024-10-14
Study data: include limit lab data in table


aclark02-arcus:
Hi @LDSamson, what do you think about adding additional data here so that users can see the upper & lower limits for the lab values? Just thinking it'd be nice to see just how far away a patient may be from the limit, if out of bounds. For example. the BMI value below is technically out of limits, but you wouldn't know that by looking at the table.

<img width="913" alt="image" src="https://github.com/user-attachments/assets/1d4aae1f-5995-491a-9e97-dac37e22d62d">

I realize that may make the table a little messier, so I'm open to alternative ways of displaying the bounds.

LDSamson:
I think on-hover labels would probably be nice. Not sure if that would be feasible though. It would be nice to add these limits to the on-hover labels of the figures as well. I felt that I was missing this information when doing reviews myself.

We can also add a color for each cell to show whether a value is out of bounds. 

Or we could do both.

LDSamson:
Related/same issue: #9

LDSamson:
It can be a toggle as well: show limits in table/don't show limits in table/show limits in separate table on the side

aclark02-arcus:
> It can be a toggle as well: show limits in table/don't show limits in table/show limits in separate table on the side

@LDSamson, I forgot to ask about this yesterday. In the demo you showed, are lab limits visible in the table view? That would be great.

LDSamson:
> > It can be a toggle as well: show limits in table/don't show limits in table/show limits in separate table on the side
> 
> [@LDSamson](https://github.com/LDSamson), I forgot to ask about this yesterday. In the demo you showed, are lab limits visible in the table view? That would be great.

Not yet, I think that should be a separate PR. 
It can be implemented with a similar switch/toggle for the tables though I think

PR #249, opened 2025-11-23
Improve figure flexibility

### LINKED ISSUE #242: Data from before baseline event are not shown in figures if baseline event is set to day zero
Usually, the day of first treatment is regarded as day zero. The day of screening can therefore be negative. It is possible to set this since https://github.com/openpharma/clinsight/pull/217, by setting the baseline event to something else than screening: 

Image

However, negative days currently do not appear in the interactive graphs and therefore screening values can be missed unintentionally if this setting is used.

LDSamson:
Fixes #242 , and in addition brings more flexibility to the figures. 

**Updated 01 Dec 2025**

Changes:

1) Now also days before baseline will be shown; axis will not be cut off anymore, which fixes #242
2) By default, only patterns of the selected patient are now shown. This should be faster to load for large data sets. Also, the pattern over time is better to see for the single patient, especially if time series of other patients are in different units.
2) There is now a new toggle in the side panel to enable/disable the background patterns of other patients for comparison
3) If background patterns of other patients are shown: the on-hover labels of other patients can now also be turned on or off
4) Within the on-hover labels, now also lab limits and the investigators judgement will be shown (Related issues: #110 and #9).

[more_fig_settings](https://github.com/user-attachments/assets/b9408f75-2654-4bda-a7b3-b4ec8d505693)


jthompson-arcus:
@LDSamson having the default only showing the one patient is preferred on our end as well.

Also, can the toggle have the same wording as the option for the tables "Show all participants"?

R/mod_study_forms.R
```diff
@@ -129,6 +147,15 @@ mod_study_forms_server <- function(
...
     }
+    observeEvent(input$show_all_participants, {
+      if(isFALSE(input$show_all_participants)){
+        shinyWidgets::updateMaterialSwitch(
+          session = session,
+          inputId = "show_all_hover_labels",
+          value = FALSE
+        )
+      }
+    })
```
jthompson-arcus: What is the point of this?

jthompson-arcus (commented):
Just one comment

R/mod_study_forms.R
LDSamson: Can be removed indeed, I think. The hover labels will only be added it both `show_all_participants` and `show_all_hover_labels` are TRUE so this is a bit double.

LDSamson: Removed in [426c8cb](https://github.com/openpharma/clinsight/pull/249/commits/426c8cb7506c72843e2fd603f484c5c325948ef8)

jthompson-arcus (approved):
LGTM

Issue #224, opened 2025-06-17
Put limits on column widths


aclark02-arcus:
When columns contain values that are really long, the `{DT}` tables are currently configured to display them without text wrap. That approach takes up lots of valuable real estate on the screen and causes for excess horizontal scrolling. It would be worthwhile to consider putting a character limit on the width of any column. However, maybe `Name` columns could get preferential treatment? It's up for debate.

![Image](https://github.com/user-attachments/assets/15fd46c3-a29a-4669-8b80-56cb9af137e2)

LDSamson:
I agree that this can be less than ideal. Do you know if there is any way to enable text wrap again?

LDSamson:
This issue is troublesome for some of our tables. For example, there is a free text field in the adverse events table that can have a lot of text which is quite hard to read on one line.

I would like to introduce a toggle to turn off deferred rendering. It will be on by default but then the user can choose to turn it off, which enables multi-line tables. 

@Jeff-Thompson12 do you see any problem with this? I think most of the time when you look at a table of just one person, this should not be a big issue, right? It would be a fairly easy solution for this problem and provides some flexibility for the user to choose table viewing of their liking.

PR #252, opened 2025-12-09
Add table text wrap toggle

LDSamson:
Mostly solves #224 .

Provides the use to enable text wrapping. The option is off by default, and a warning is shown that it might be slow to enable it for big tables

grafik


R/mod_common_forms.R
```diff
@@ -148,6 +169,7 @@ mod_common_forms_server <- function(
         form_items = form_items,
         active_subject = active_subject,
         show_all = reactive(isTRUE(input$show_all_data) | identical(session$userData$review_type(), "form") ),
+        enable_text_wrap = reactive(isTRUE(input$enable_text_wrap)),
```
jthompson-arcus, line 152:
```suggestion
        enable_text_wrap = reactive(isTRUE(input$enable_text_wrap) & identical(session$userData$review_type(), "subject")),
```
For consistency

R/mod_study_forms.R
```diff
@@ -176,6 +200,7 @@ mod_study_forms_server <- function(
       active_subject = active_subject,
       form_items = form_items,
       show_all = reactive(isTRUE(input$show_all) | identical(session$userData$review_type(), "form")), 
+      enable_text_wrap = reactive(isTRUE(input$enable_text_wrap)),
```
jthompson-arcus, line 285:
```suggestion
      enable_text_wrap = reactive(isTRUE(input$enable_text_wrap) & identical(session$userData$review_type(), "subject")),
```
For consistency

jthompson-arcus (changes requested):
Enabling text wrap does not appear to be working. For instance checkout subject BEL_04_772 and the Medication tab. There are 15 records. If you "enable text wrap" you can only view 10 of them.

jthompson-arcus:
I think generally I'm a bit confused by this PR. In the common events the text wrapping is disabled on form level review. I assume this is to mitigate issues from having all records visible but the similar checking doesn't seem to be in place for the toggle to see all subjects. The toggle doesn't disappear at the form level either so it all feels a bit messy to me.

LDSamson:
> I think generally I'm a bit confused by this PR. In the common events the text wrapping is disabled on form level review. I assume this is to mitigate issues from having all records visible but the similar checking doesn't seem to be in place for the toggle to see all subjects. The toggle doesn't disappear at the form level either so it all feels a bit messy to me.

Hi, thanks for the review. Initially, the idea was to discourage to use it in form level review, but not to disallow it. Do you think it should be completely forbidden?

jthompson-arcus:
> > I think generally I'm a bit confused by this PR. In the common events the text wrapping is disabled on form level review. I assume this is to mitigate issues from having all records visible but the similar checking doesn't seem to be in place for the toggle to see all subjects. The toggle doesn't disappear at the form level either so it all feels a bit messy to me.
> 
> Hi, thanks for the review. Initially, the idea was to discourage to use it in form level review, but not to disallow it. Do you think it should be completely forbidden?

If this feature is something desired, I personally would think about turning pagination back on if it is triggered. You are re-rendering the table anyways.

LDSamson:
> Enabling text wrap does not appear to be working. For instance checkout subject BEL_04_772 and the Medication tab. There are 15 records. If you "enable text wrap" you can only view 10 of them.

@jthompson-arcus Text wrap works it's just that not all entries are shown indeed which I did not expect.. do you know why this is the case?

I will try to switch to pagination again and see if that works.

LDSamson:
> The toggle doesn't disappear at the form level either so it all feels a bit messy to me.

I am a bit confused by this comment.  The toggle should disappear when using form_level review. Is this not the case for you? 
I updated the PR, pagination is now again enabled when text_wrap is enabled. Let me know what you think! 

text wrap can indeed still be enabled when pressing show_all_participants but not when using form_level review, I thought for the latter it is probably not needed.

jthompson-arcus (changes requested):
I still do not understand why we would not want the end user to be able to enable wrapping when reviewing by form. The word wrap causes the table to display paginated, so we should allow that view for form level reviews as well.

Also, perhaps padding between the AE tables should be considered.

LDSamson:
Text wrap is now consistent between form-level review and just pressing the `show all participants` button. It can be used with both. 

It switches to no text wrapping when clicking on 'show all participants'. This is because in this mode, it is probably easier to compare lots of data if the data is on one row. But you can still switch on text wrap if desired.

I also added extra padding under the SAE table as suggested and removed the tooltip warning since it seems like text wrap is still okay for large datasets when pagination is enabled (which is the case here)

jthompson-arcus (changes requested):
I know this is turning into the PR of a thousand change requests but...I found a bug.

If you select rows to review and then toggle "enable text wrapping", the rows will still be cached and you can hit the "save" button but the rows in the table do not show being selected.

LDSamson:
> I know this is turning into the PR of a thousand change requests but...I found a bug.
> 
> If you select rows to review and then toggle "enable text wrapping", the rows will still be cached and you can hit the "save" button but the rows in the table do not show being selected.

@jthompson-arcus thanks for spotting this. Now resolved in this commit: 2c0a2e14f1779abd0ea118d3d35d8886d8ad67da.
I also added a test to ensure that it is resolved: 2c0a2e14f1779abd0ea118d3d35d8886d8ad67da

R/mod_review_form_tbl.R
```diff
@@ -149,9 +154,9 @@ mod_review_form_tbl_server <- function(
       table_data(df)
     })
     
-    # Any time the data in the form table is updated, "show all" is toggled,
-    # or the subject being viewed is changed, the server data for the datatable
-    # needs to be updated
+    # Triggers when server data needs to be updated. Also triggers for each 
+    # change in pending review records (e.g. a checkbox in column `Reviewed` 
+    # is toggled on or off).
```
LDSamson, line 159: @jthompson-arcus changed the description here and in lines 173-174 a bit to capture the 'why' because it always take me some time to familiarize with the logic here after a while. If you don't like it or have suggestions for better description please let me know.

jthompson-arcus (approved):
Looks good to me now. Thanks for the updates.

PR #253, opened 2025-12-09
Column name line clamping

jthompson-arcus:
Apply line clamping to long column names that go over multiple lines.


jthompson-arcus:
@LDSamson a little birdie told me that there may be interest in adding this feature to handle tables with long column names.

R/fct_data_helpers.R
```diff
@@ -529,6 +529,16 @@ datatable_custom <- function(
     stopifnot(is.character(rename_vars))
     colnames <- dplyr::rename(data[0,], dplyr::any_of(rename_vars)) |> 
       names()
+  }  
+  if ("escape" %in% ...names()) {
+    if (isFALSE(...elt(match("escape", ...names()))))
```
LDSamson: Isn't it much easier to just explicitly define `escape` in `datatable_custom`, with a default value of `TRUE`?

Then this statement can just be:

```
if (isFALSE(escape)) {
  ...etc...
}
```

R/fct_data_helpers.R
```diff
@@ -529,6 +529,16 @@ datatable_custom <- function(
...
+      colnames <- purrr::map_chr(
+        colnames, 
+        \(cn) as.character(tags$span(
+          cn, 
+          class = "cs-span-overflow",
+          title = cn
+          )))
```
LDSamson: Suggestion to change to this:

```
    lapply(
      colnames, 
      \(cn) { 
        as.character(tags$span(
          htmlEscape(cn), 
          class = "cs-span-overflow",
          title = cn
        ))
      }
    ) |> 
      as.character()
```

Reasoning: 

- `purrr` is currently only in the `Suggest` field in the description. We should either use base R or move  `purrr` in the description to `Imports`.
- Using `htmlEscape` might be overly protective but on the other hand it probably won't hurt here.

LDSamson: Suggestion to change to this:

```
    lapply(
      colnames, 
      \(cn) { 
        as.character(tags$span(
          htmlEscape(cn), 
          class = "cs-span-overflow",
          title = cn
        ))
      }
    ) |> 
      as.character()
```

Reasoning: 

- `purrr` is currently only in the `Suggest` field in the description. We should either use base R or move  `purrr` in the description to `Imports`.
- Using `htmlEscape` might be overly protective but on the other hand it probably won't hurt here.

LDSamson (commented):
Hi,

thanks for this addition, I like it. Just a few suggestions here, see comments

R/fct_data_helpers.R
jthompson-arcus: Just copied and pasted from our implementation. Chose this approach at the time because I did not want to change arguments for functions.

jthompson-arcus: Resolved in a9b03ae

R/fct_data_helpers.R
jthompson-arcus: Resolved in fadb525

LDSamson (approved):
LGTM!

PR #255, opened 2025-12-12
Additional navigation btns

LDSamson:
Added some simple new navigation buttons in the tabsets.

Reason is that sometimes people forget that there is also study data when viewing common events.
In addition, it makes the navigation easier

https://github.com/user-attachments/assets/29ce7e98-31d2-42f0-b4ae-497885fd5a21


aclark02-arcus (approved):
Looks great. I ran the app locally & it works well. A good add!

PR #256, opened 2025-12-15
Add query role privilege

### LINKED ISSUE #17: Queries: implement role-based authority for closing queries.
For example: 
- MM can close all queries 
- Data Manager can close Normal queries.

Maybe set these options in golem-config.yml (or a new, separate config file that can be provided at runtime?)

LDSamson:
Closes #17.

- Adds more role privileges (e.g. who can save a query).
    - Hides query button if the role does not allow to query
    - Everyone with a role assigned can write follow-up messages to a query. However, people without the `allow_to_query()` role cannot close queries.
- Polished some of the messages regarding role-specific privileges. 
    - For example, if no role is assigned, the message is less dramatic and does not sound like there is a major error anymore -> this is not needed. In ShinyProxy setups, it can occur that people do not have a role assigned but has access to the app; then the person simply has viewing acces.
- Makes it easier to add a clinsight custom configuration by searching in the working directory for `clinsight-config.yml` and using that before falling back to the default `golem-config.yml` file in the package


R/fct_utils.R
```diff
@@ -790,7 +790,15 @@ expectation_type <- function(
 #' 
 custom_config_path <- function(
 ){
-  Sys.getenv("CONFIG_PATH", app_sys("golem-config.yml")) 
+  config_path <- Sys.getenv("CONFIG_PATH")
+  if (file.exists(config_path)) {
+    return(config_path)
+  }
+  if (file.exists("clinsight-config.yml")) {
+    "clinsight-config.yml"
```
jthompson-arcus, line 798: This change should be documented in the Deployment vignette.

R/mod_query_follow_up.R
```diff
@@ -63,9 +63,17 @@ mod_query_follow_up_server <- function(id, r, selected_query, db_path){
...
+        shinyjs::enable("query_follow_up")
       }
     })
+    
+    observeEvent(r$user_role, {
+      shinyjs::toggleElement(
+        "resolved", 
+        condition = get_roles_from_config()[r$user_role] %in% get_golem_config("allow_to_query")
+      )
+    })
```
jthompson-arcus: This only causes the checkbox to not be displayed. A user could in their browser make the checkbox visible and still "resolve" a query even without the permissions.

R/mod_query_follow_up.R
jthompson-arcus, line 1: Actually realizing there's not a ton a server side verification being done here. I'm also able to comment on "resolved" queries as well just by removing the disabled attribute in the browser.

jthompson-arcus (changes requested):
As I understand it, this role is meant to restrict query creation and resolution. But any user can still respond to queries correct?

R/mod_query_follow_up.R
LDSamson: Thanks for spotting this. Resolved in [361daf0](https://github.com/openpharma/clinsight/pull/256/commits/361daf0f1bc4f54e484a8631594b1a7365e35f70)

R/mod_query_follow_up.R
LDSamson: Resolved in [e95f72c](https://github.com/openpharma/clinsight/pull/256/commits/e95f72cdbde69f6a77d22e2d5e05a2fabee02076)

R/fct_utils.R
LDSamson: Resolved in [64497c1](https://github.com/openpharma/clinsight/pull/256/commits/64497c167e5c8ad15a292b9430dbd8030febddef). Also updated the clinsight config helper to make it consistent: [da33777](https://github.com/openpharma/clinsight/pull/256/commits/da33777e47e014714191fac9639959c09dfb6b26).

jthompson-arcus (approved):
LGTM

Issue #221, opened 2025-06-17
Timeline - improvement ideas


aclark02-arcus:
1. Make separate treatments easily identifiable - show name of drug and different icon &/or color.
2. Left justify & add directional arrows for AEs that are ongoing (they have no end date)
3. Show intended dose dates in timeline (when available), so we know when a dose was missed, and the users could visually see & maybe conclude it was due to a recent AE
4. If there is a start & end date for a dose (ex. daily oral capsule from date1 to date2), the dose treatment should be represented as a bar. When the dose stops, there should be a gap.
5. Attempt to "link" Con Meds with Adverse Events & show them in the timeline. 
    - In visual AND in table
    - Separate but related,  attempt to link Con Meds & Medical History

aclark02-arcus:
With #251, this will be easier to work on!

aclark02-arcus:
@LDSamson, just added 3 & 4 above. Do any of those situations apply to you at GCP? Just curious

Issue #12, opened 2024-05-22
Timeline: Add support within timeline to display any data. 


LDSamson:
For example, create a modal with timeline, with option to add/remove common form data, and add values of study data on the bottom (or maybe graphs under the timeline?).

Or, add a possibility to add timeline data/ events in the graphs?

PR #258, opened 2025-12-16
Adjust timeline

LDSamson:
PR is part of the process of adding more items in the timeline, such as concomitant medication and/or concomitant procedures. 
Will help to solve this: #12 and some of this: #221

Two main changes here: 

- Places the timeline in the header instead of in mod_common forms
- Adds a toggle to optionally show the timeline

Timeline is now on by default when viewing common forms and off when viewing study forms.

https://github.com/user-attachments/assets/6947dc67-a46c-47a5-ad02-c51ab95360b1


jthompson-arcus (commented):
Mostly focusing on visuals for this review. Didn't have time to dig into the code.

- At first I thought the toggle should be above the timeline but I think it makes the most sense to have it as you do as part of the nav bar. I do feel like the toggle is slightly too big and feels awkward.
- I find turning the timeline off and on a bit jarring. I think some transitions need to be added but that's a bit more involved.

@aclark02-arcus what do you think?

aclark02-arcus:
Agree with Jeff's comments & adding one more. I don't think the toggle is too big... but I don't think it's obvious what it does just by looking at it:

image

It's really hard to tell what that icon is displaying. Would it be too crazy to just include a "Timeline" label next to the toggle?

LDSamson:
Thanks for the suggestions. I added an animation and changed the toggle, see changes below. 
@aclark02-arcus and @jthompson-arcus what do you think?
https://github.com/user-attachments/assets/2892a88f-e466-4fbd-84d8-7bbbf35209bc

aclark02-arcus:
@LDSamson, that looks real nice. I only hope people don't confuse it with the static visit timeline above.

jthompson-arcus (commented):
This looks good to me. Just need to sync up with the `dev` branch.

jthompson-arcus (approved):
LGTM

PR #259, opened 2025-12-17
Add data transformation selection

### LINKED ISSUE #110: Study data: include limit lab data in table

### LINKED ISSUE #261: Feature - Choose between data transformations
There are a few features that have priority for us to implement, mostly about better managing laboratory data from different sites with different units. 

- Right now limits are only visible when scaling figures. 
- Furthermore, if limits are missing and the scaled option is chosen, the values do not appear in the figure.
- There is a method missing to convert units to a single standard unit. 
   - This can first be out of scope of clinsight to calculate this. But ideally it should be optional to provide visualizations or tables with these values

Proposed solutions:
- Create a toggle to switch between data transformations. 
    - no transformation
    - scaled transformation
    - If a column with this data exists: standardized unit transformation

The local lab range should also be visible in the view with no transformation.

Both the `no transformation` and   `standardized unit transformation`  should be available in table view and in figure view.
Ideally also a warning should be shown if not all values are available in one of the scaled views, and advise to view the raw data.

Implementing it this way, making the veiw optional and with the warning, the visibility of a possible error will be high and therefore will reduce the risk class of this feature.

LDSamson:
Closes #261 and #110


aclark02-arcus (approved):
@LDSamson, this looks great! Can't wait to try it out on our internal deployment. 🔥 
Just need to fix R CMD Check.

tests/testthat/_snaps/app_feature_01/app-feature-1-003.json
```diff
@@ -947,279 +947,11 @@
...
-              -5,
-              -5,
-              -5,
-              -5,
-              -5,
-              -5,
-              -5,
-              -5,
-              -5,
-              -5,
```
LDSamson, line 961: @aclark02-arcus  Of note is that one lab limit per data point was drawn previously. This was not shown because the limits overlapped, see here in this snapshot. This is now corrected, which will help with the speed of the figures as well (and it explains the dramatic removal in code lines, see below  😆 )

Image

PR #266, opened 2026-01-12
Fix error in figures if no data is available for selected subject and `show_all_participants` is `FALSE`

LDSamson:
To reproduce, browse to subject `NL_05_282` in the current `dev` version and view the `Electrolytes`  tab. The following error shows up:

grafik

This fix is easy, just one additional line and it looks again as desired:
grafik

I did not add a news item since this bug was introduced in PR #249, and thus only existed in the `dev` version. Happy to add some if you think it is needed, no strong feelings about that

GAMP V Low-risk priority because it is a non-fatal error, clearly visible, and usability is not compromised.

### DIFF
--- a/R/fct_figures.R
+++ b/R/fct_figures.R
@@ -234,7 +234,7 @@ fig_timeseries <- function(
     ggplot2::facet_wrap(~item_name, ncol = 2, scales = "free_y") +
     ggplot2::scale_fill_manual(values = col_palette) +
     ggplot2::scale_x_continuous(limits = \(x){
-      if(length(x) == 0) return(c(0,1))
+      if(length(x) == 0) return(c(0,3))
       c(
         pmin(x[1], 0), # Always include day zero. 
         pmax(x[2], 3) # keeps minimum scale of 3 days if not much data is available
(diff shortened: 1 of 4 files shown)


R/fct_figures.R
```diff
@@ -234,6 +234,7 @@ fig_timeseries <- function(
     ggplot2::facet_wrap(~item_name, ncol = 2, scales = "free_y") +
     ggplot2::scale_fill_manual(values = col_palette) +
     ggplot2::scale_x_continuous(limits = \(x){
+      if(length(x) == 0) return(c(0,1))
```
jthompson-arcus: Very small comment but why not `c(0,3)` which is more inline with `pmax(x[2], 3)`?

LDSamson: Yes that is more consistent indeed. No specific reason. I will change it to `c(0.3)`

LDSamson: Resolved in [a5fab16](https://github.com/openpharma/clinsight/pull/266/commits/a5fab162746ad72f3110ba1041b07b685a399dff)

jthompson-arcus (approved):
LGTM

PR #268, opened 2026-01-30
More consistent toggles

LDSamson:
Purely visual update to make the toggles more consistent with each other and giving clearer, shorter labels with some icons for clarity.


aclark02-arcus (approved):
🔥

PR #269, opened 2026-02-03
Add `origin` arg to `as.Date()` in `get_timeline_data()`

aclark02-arcus:
Posit Connect was throwing a fit with this line of code for some reason. Issue resolve with the following change.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: clinsight
 Title: ClinSight
-Version: 0.3.0.9014
+Version: 0.3.0.9015
 Authors@R: c(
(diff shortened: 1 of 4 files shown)


LDSamson (changes requested):
At a first glance looks like a harmless change. 
Any idea why the test is failing though? Can you have a look at that?

Furthermore, do you have any idea why this is needed? I';m a bit puzzled about the why in this casse. Is you date variable not a character vector before conversion, maybe it is somethign like that?

LDSamson (approved):
@aclark02-arcus I added some wait time in the test to make it more robust. 
Otherwise, looks good to me. I will bump version and merge. 

We need to refactor this function though, the core issue is that base R `ifelse` strips date classes. In addition, it can be simplified so that events from other forms can be added easily. But I think we should do that in a separate PR for the next release.

PR #275, opened 2026-03-03
Deploy `v0.4.0` to shinyapps.io

aclark02-arcus:
### Steps
- Had to install `utils::install.packages("rsconnect")` for some reason
- In `/app.R`, update `manifest.json`... though I'm not convinced this is needed for shinyapps.io... I think it relies on the renv.lock file only. But this is nice to have for a deployment to Posit Connect.
- Moved full `renv.lock` to root dir

### Deployments
- https://rinpharma.shinyapps.io/clinsight_v040/
- https://rinpharma.shinyapps.io/clinsight/


aclark02-arcus:
@LDSamson, deployments are prepared! I'll just leave this PR as draft for now, just in case. But I assume there is no need to merge this branch.

LDSamson:
I checked and it works and is online indeed. Thanks!

